Issue #33, opened 2015-02-14
VWLineSimplifier Optimization


airbreather:
Not exactly sure where I should be putting this, as I don't think it's worthy of a pull request just yet, but I've got a heavily optimized version of VWLineSimplifier that I'm tentatively calling "FastVWLineSimplifier".  The TODO caught my eye and I couldn't resist.

airbreather/NetTopologySuite@ee14fc978516541727de39c701389bc59f4130f5

That commit includes a test case that produces output similar to this on my machine:
Time (slow): 11.9271277033042 seconds
Time (fast): 0.0595896480771026 seconds

Notably, FastVWLineSimplifier appears to return perfectly identical results to what VWLineSimplifier would produce (doing a full replacement passed all the NUnit tests), though the implementation is obviously completely different.

I think all it should really need is some surface-level cleanup and general direction from you guys about what to do about it.

Thoughts?

FObermaier:
Great! As we try to keep NTS in line with JTS I think we should inform @dr-jts about your improvement.

DGuidi:
wow, thanks :)
to me, we should already integrate code into NTS, maybe in an "experimental" namespace, and then, if and when @dr-jts integrates code into JTS, adapt NTS code to the JTS one.

DGuidi:
@airbreather feel free to open a pull request when you think code is stable enough

dr-jts:
Nice...  

JTS already has a PriorityQueue:

http://sourceforge.net/p/jts-topo-suite/code/HEAD/tree/trunk/jts/java/src/com/vividsolutions/jts/util/PriorityQueue.java

Is that in NTS, and if so would it work for this?

airbreather:
@dr-jts: It [appears so](https://github.com/NetTopologySuite/NetTopologySuite/blob/master/NetTopologySuite/Utilities/PriorityQueue.cs); the main thing I wanted to solve before making a proper pull request is to find the priority queue implementation that made the most sense to use.  This may be it.

FObermaier:
@airbreather, would you try to adopt your code to using the PriorityQueue in the NetTopologySuite.Utilities namespace?

airbreather:
@FObermaier yeah, working on it now.

airbreather:
[...]

PR #34, opened 2015-03-14
Optimizing VWLineSimplifier Using a Priority Queue

airbreather:
This is what we discussed in #33.

Some notes:

1. AlternativePriorityQueue<TPriority, TData> is _derived from_ BlueRaja's implementation, but it makes a lot of tiny speed tradeoffs in exchange for reliability and user-friendliness.
2. 3036ada is completely optional.  I think it will have the tendency to perform worse than what's there now, but it does get rid of a lot of now-redundant code.
3. PriorityQueueTest.cs changes seem to make it better than what's there now.  Before, CheckOrder would do a lot of boxing / unboxing, and all it would do is just make sure that everything after the first Poll() is greater than or equal to what the first Poll() returned.  Now, it makes sure that all values are returned in the correct order relative to one another.
4. TestNewResultsIdenticalToOldResults uses what I eyeballed to be the data from the TestData folder that should most clearly demonstrate the execution time benefits.  The new version is an 86% execution time reduction from the old version.  The new version will do even better relative to the old version when the input lines are much longer than these.
5. Since this queue implementation is derived from BlueRaja's implementation, I think versions of NTS based on this would need to distribute some extra license file to ensure compliance.  I didn't do anything here in consideration of that, I figured you guys would be able to work it out easily enough since you're distributing a modified PowerCollections as well, which has what looks like a less permissive license.

Have at it!

### DIFF
--- a/NetTopologySuite.Tests.NUnit/NetTopologySuite.Tests.NUnit.csproj
+++ b/NetTopologySuite.Tests.NUnit/NetTopologySuite.Tests.NUnit.csproj
@@ -210,6 +210,7 @@
     <Compile Include="Triangulate\DelaunayPerformanceTest.cs" />
     <Compile Include="Triangulate\DelaunayTest.cs" />
     <Compile Include="Triangulate\VoronoiTest.cs" />
+    <Compile Include="Utilities\AlternativePriorityQueueTest.cs" />
     <Compile Include="Utilities\PriorityQueueTest.cs" />
     <Compile Include="Utilities\SerializationUtility.cs" />
   </ItemGroup>
--- a/NetTopologySuite.Tests.NUnit/Simplify/VWSimplifierTest.cs
+++ b/NetTopologySuite.Tests.NUnit/Simplify/VWSimplifierTest.cs
@@ -1,7 +1,13 @@
 using System;
+using System.Linq;
+
 using GeoAPI.Geometries;
+
+using NetTopologySuite.Geometries.Utilities;
 using NetTopologySuite.IO;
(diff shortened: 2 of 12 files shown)


DGuidi:
@FObermaier what do you think about?

FObermaier:
I'm in favor of accepting the pull-request.
I'd say we mark the non-JTS additions as such. If JTS adopts these changes or implements it differently. We can follow easily.

@drjts: any news on the license change for JTS?

FObermaier:
@dr-jts: any news on the license change for JTS?

dr-jts:
Working on it...

DGuidi:
Ok for me. We should modify also license file. You should merge the pull request if you want. In a couple of days I can accomplish the task, anyway.

----- Messaggio originale -----
Da: "Felix Obermaier" <notifications@github.com>
Inviato: ‎18/‎03/‎2015 17:57
A: "NetTopologySuite/NetTopologySuite" <NetTopologySuite@noreply.github.com>
Cc: "Diego Guidi" <diegoguidi@gmail.com>
Oggetto: Re: [NetTopologySuite] Optimizing VWLineSimplifier Using a PriorityQueue (#34)

I'm in favor of accepting the pull-request.
I'd say we mark the non-JTS additions as such. If JTS adopts these changes or implements it differently. We can follow easily.
@drjts: any news on the license change for JTS?

FObermaier:
Thanks @airbreather!

DGuidi:
@airbreather thanks, again!

airbreather:
No problem, glad I could help.

PR #48, opened 2015-09-01
Allows shapefiles to be read from arbitary streams, and adds an implementation using windows azure cloud storage

jaundice:
I have modified the shapefile.extended to allow the use of arbitary streams. It defaults to using FileStream and all dependent existing code should work as is. There is also a new project and associated tests implementing shapefile reading from windows azure cloud storage


jaundice:
@FObermaier  hold this. I have further changes to make

jaundice:
@FObermaier  I have made further changes - moving the stream provision out to a separate project, then consuming it in geotools as well as shapefile.extended. all tests pass. Unfortunately there are quite a lot of whitespace changes introduced as well.. My R# reflexes get the better of me!

FObermaier:
Thanks John, I have three suggestions
1. I would be to put the contents of NetTopologySuite.IO.Common.Streams into plain NetTopologySuite.IO project (I can do that)
2. Add `string Kind { get: }` readonly property to IStreamProvider
3. Change the `ICombinedStreamProvider` interface as follows:
`
interface IStreamProviderRegistry 
{
    IStreamProvider this[string fileType] { get; }
}
`
and remove all the specific IStreamProvider interfaces.

jaundice:
Sounds good to me.

DGuidi:
@jaundice @FObermaier some code contains String.IsNullOrWhiteSpace calls, that are specific of Framework 4.0. I've tried to build the whole framework using msbuild and I've experienced some errors.

jaundice:
@DGuidi  see my latest pull request. Hopefully that issue is fixed...

DGuidi:
@jaundice thanks for the fix

PR #50, opened 2015-09-03
replaced string.IsNullOrWhitespace

jaundice:
@DGuidi @FObermaier  this dispenses with string.IsNullOrWhitespace by adding an extension method in Utilities.StringEx hopefully that fixes the MSBuild issue


FObermaier:
@jaundice: Please update your PowerCollections subrepository

jaundice:
should be ok now

PR #66, opened 2015-11-08
Decently write big long and ulong attribute values to shapefiles.

xivk:
Decently write big long and ulong attribute values to shapefiles. The current implementation limits integer values to maximum of 10 chars.


xivk:
Thanks for the quick review & merge!

PR #70, opened 2015-12-25
Optimizing the big-endian binary reader/writer.

airbreather:
Currently, BEBinaryReader and BEBinaryWriter do byte order reversing for each value read (edit: or, of course, written) by allocating a new byte array on the heap, reversing the bytes in that array, and then using BitConverter to interpret the data in that array.  At least on my machine, doing this extra intermediate array allocation for every single value we read is costly (according to some measurements).

The way [JTS](http://sourceforge.net/p/jts-topo-suite/code/1118/tree/trunk/jts/java/src/com/vividsolutions/jts/io/ByteOrderDataInStream.java) deals with this [problem](http://sourceforge.net/p/jts-topo-suite/code/1118/tree/trunk/jts/java/src/com/vividsolutions/jts/io/ByteOrderValues.java) is each reader and [writer](http://sourceforge.net/p/jts-topo-suite/code/1118/tree/trunk/jts/java/src/com/vividsolutions/jts/io/WKBWriter.java) allocates a fixed number of byte buffers so that there's a fixed number of byte array allocations throughout the lifetime of a particular reader instead of one per value read.

The approach I took was to just create helper methods that reverse the byte order of the primitive value type given to it using regular ol' shifting and masking... for all but `float` / `System.Single`, which would require either `BitConverter.SingleToInt32Bits` / `BitConverter.Int32BitsToSingle` such as what coreclr pull request [#833](https://github.com/dotnet/coreclr/pull/833) would add, or turning off the guard to allow unsafe blocks and reimplementing the same.  Since I don't think anything in NTS currently uses the `float` methods, I think just continuing to do what we were doing before should be fine.

### DIFF
--- a/NetTopologySuite/IO/BEBinaryReader.cs
+++ b/NetTopologySuite/IO/BEBinaryReader.cs
@@ -1,8 +1,8 @@
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Text;
 using GeoAPI.IO;
+using NetTopologySuite.Utilities;
 
 namespace NetTopologySuite.IO
 {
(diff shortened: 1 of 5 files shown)


DGuidi:
@FObermaier +1 to merge this

PR #71, opened 2016-01-16
Optimizing PointLocator a bit.

airbreather:
Now tests with the underlying ICoordinateSequence, which is much nicer to the GC for sequence types other than CoordinateArraySequence (for which there's effectively no change).
Also hoisting the ICoordinateSequence.Count property accessor outside of the loop.

I used this for testing: https://gist.github.com/airbreather/72a044104a4d44014235 (follow the link in the source for where the testing data came from).
That `ICoordinateSequence` implementation is defined over here if you really need to see it (you probably don't): https://gist.github.com/airbreather/d227a19c74799c6b45e4

Before: 188.4 seconds.
After: 64.5 seconds.

I only did one run because I'm impatient.  I probably should have done a few and averaged, but I made an educated guess that the message isn't going to change by doing so.

### DIFF
--- a/NetTopologySuite.Tests.NUnit/Algorithm/LocatePointInRingTest.cs
+++ b/NetTopologySuite.Tests.NUnit/Algorithm/LocatePointInRingTest.cs
@@ -16,7 +16,11 @@ namespace NetTopologySuite.Tests.NUnit.Algorithm
         {
             IGeometry geom = reader.Read(wkt);
             Assert.AreEqual(expectedLoc, CGAlgorithms.LocatePointInRing(pt, geom.Coordinates));
+            IPolygon poly = geom as IPolygon;
+            if (poly == null)
+                return;
 
+            Assert.AreEqual(expectedLoc, CGAlgorithms.LocatePointInRing(pt, poly.ExteriorRing.CoordinateSequence));
         }
 
     }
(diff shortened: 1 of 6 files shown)


airbreather:
I've been running a bigger version of the test all day for a second PR I'm going to submit once this one goes somewhere... along the way I got some additional numbers for running on `WholeCountry` from that gist (a much bigger test).

After: 5433.7 seconds (90.6 minutes)
Before: 20700 seconds (345 minutes) **and counting** (it's been running all day).

DGuidi:
awesome! looks something that can be suggested to martin for JTS, too?

airbreather:
@DGuidi: it would [appear](http://sourceforge.net/p/jts-topo-suite/code/1125/tree/trunk/jts/java/src/com/vividsolutions/jts/algorithm/PointLocator.java#l190) that way.

@dr-jts: if I mention your name here, do you see this?  Summary is that when I added `CoordinateSequence` overloads to the `*PointInRing` methods in `CGAlgorithms` and had the equivalent of `PointLocator.locateInPolygonRing` pass in the `CoordinateSequence` on the `LinearRing`, it made `PointLocator` significantly faster for point-in-poly checks on geometries created from packed coordinate sequences, etc.

Yes, I know about the faster alternative(s) to `PointLocator`, that's my next PR.

DGuidi:
merged, thanks.

PR #74, opened 2016-01-23
Removing PowerCollections

airbreather:
The main reason I'm pushing this to get the build server to run the unit tests on it.  You can look at it if you want, but I'm not finished doing my own self-review of it yet.

Issue #73, opened 2016-01-22
[Question / Discussion] PowerCollections in the Full .NET Framework


airbreather:
In every case but `BigList<T>`, the full .NET Framework provides a direct built-in replacement for PowerCollections types used in NTS.  Some of them ([SortedDictionary&lt;TKey, TValue&gt;](https://msdn.microsoft.com/en-us/library/f7fta44c.aspx)) only require .NET 2.0+.  Others ([HashSet&lt;T&gt;](https://msdn.microsoft.com/en-us/library/bb359438.aspx)) require .NET 3.5+.

In one case, replacing the implementation of a method from PowerCollections's `OrderedDictionary<TKey, TValue>` with [something](https://gist.github.com/airbreather/8b8544f9c4babcc490ba) based on the [implementation](https://github.com/dotnet/corefx/blob/cd235dbf9d06d64b06643007de13307bba5b92be/src/System.Collections/src/System/Collections/Generic/SortedSet.cs#L209-L244) from the built-in `SortedDictionary<TKey, TValue>` took a significant chunk off the time from an operation.  Specifically, it eliminated between 13%-14% of the total time spent in `Geometry.Union(Geometry)` from `CascadedPolygonUnion`.  I did not go in-depth to see if there were other places where we get penalized for using PowerCollections instead of a built-in.

I wonder if it would be appropriate to revisit whether or not to include PowerCollections as a dependency in targets where there's a built-in equivalent for everything?  Built-in types tend to be highly optimized.  Specifically:

1. For [QuadEdgeSubdivision](https://github.com/NetTopologySuite/NetTopologySuite/blob/0f56e478f91a0a40e4b75203ade55908990fc4ce/NetTopologySuite/Triangulate/QuadEdge/QuadEdgeSubdivision.cs), replace `ISet` with `HashSetQ` and change the `#if NET40` to `#if NET35` (`ISet` shows up once, in a private method signature that's guaranteed to always have a `HashSetQ`, so this is not a breaking change).
2. For anything that uses `OrderedDictionary<TKey, TValue>`, use `SortedDictionary<TKey, TValue>` instead, `#if NET20`.
3. For anything that uses `OrderedSet<T>`, use `SortedSet<T>` instead, `#if NET20`.
4. For anything that uses `Set<T>`, use `HashSet<T>` instead, `#if NET35`.
5. For anything that uses `BigList<T>`, re-evaluate whether or not it's actually better than `List<T>`.
[...]

PR #75, opened 2016-01-24
Removing uses of PowerCollections where possible.

airbreather:
This is related to #73.
.NET 3.5 and above can get rid of it completely.
.NET 2.0 only needs it for a hash-based set type.
PCL can even get away from some of it with simple changes.

I previously did a "dummy" PR (#74) to get the build server to test this and let me refine it so it's easier to look at.  This is the "real" PR.

Specific considerations worthy of mention:

1. `SortedSet<T>`, the direct replacement for `OrderedSet<T>` from PowerCollections, is .NET 4.0+ only.  However, everything that uses it could have used a hash-based set instead and sorted at the end (which may be faster or marginally slower than before, depending on the number of duplicates).  In one case, it didn't even care about the order, so this is a strict improvement.
    1. This is probably the most controversial code change, so I'll explain more about performance.
    2. Specifically, using `OrderedSet<T>` and then copying the results to an array is O((n + m) * log(n)), where "n" is the number of unique items and "m" is the number of items that are also found in "n".  This is because testing each element, whether it's in "m" or "n", requires an O(log(n)) lookup into the red-black tree.  Elements of "m" are not added.  Copying to an array only requires O(n) time.
    3. On the other hand, using a hash-based set, then copying the results to an array, then sorting it is O(n * log(n)) + O(n + m), for the same definitions of "n" and "m".  Now, testing each element only requires an O(1) lookup into the hash-based set, which we do (n + m) times.  So we inserted everything in O(n + m) time, then we copy everything in O(n) time (because duplicates are already excluded), then we sort in O(n * log(n)) time.
    4. For "m" large, this change is an improvement on the asymptotic execution time.  For "m" small, it's a wash.
    5. The memory layout properties of a hash-based set (especially the built-in `HashSet<T>`), compared to a red/black tree, lead me to believe that this is likely to be faster overall, but I haven't really done aggressive testing on it.
    6. **NOTE**: this change is only valid if the element has its `IComparable` / `IComparable<T>` semantics be consistent with its `Equals` / `GetHashCode` semantics such that two objects with a comparison of 0 from `CompareTo` are also considered equal by `Equals` and have identical hash codes from `GetHashCode`.  Fortunately, NTS only uses `Coordinate` as the "`T`" for its `OrderedSet<T>`, which is self-consistent in this regard.  For many `TKey` types in `OrderedDictionary<TKey, TValue>`, this is not the case, so a similar change would not be a valid way to "fix" those cases.
2. The two places that used `BigList<T>` actually seemed to just be depending on its sorting semantics (specifically, how PowerCollections determines the relative order of items that are equal).  In those places, the equivalent JTS code uses a sort that is guaranteed to be stable, so we're free to change those as long as we also do stable sort.  `OrderBy<T, TItem>` from LINQ (.NET 3.5+) is guaranteed stable.  For other targets, PowerCollections happens to have a `StableSort<T, TItem>` available to us with nearly identical semantics.
    1. For what it's worth, `BigList<T>` claims to use QuickSort, which is typically not stable.  I did not continue down this road to see whether or not there's actually a bug that this change fixes.
3. The default build for NetTopologySuite didn't define `NET40`, even though it builds for .NET 4 Client.
4. Along with removing the PowerCollections target from the builds where it's not needed, we also need to have that "NeedsPowerCollections" conditional to avoid the PowerCollections stuff getting copied to the output directory anyway.

For the future, here are the things keeping .NET 2.0 and PCL from dropping it entirely:

1. We need a hash-based set type.  It's possible to emulate `HashSet<T>` by wrapping a `Dictionary<T, bool>`.  I didn't want to do this right away.
2. We need a way to do a stable-sort.  One way to do it would be to add a utility method that wraps each element of an `IEnumerable<T>` in an `ItemWithIndex<T>` whose comparison implementation uses the original `T`'s comparison, and then uses the index from the original sequence as a tiebreaker, then sort it and select the `T` elements of the sorted set back out.
    1. Of course, another way would be to just go ahead and implement a stable sort within NTS, which is not very appealing either.
3. We need an ordered dictionary type.  As mentioned above, it's not actually all that easy to do this.  I have some ideas, but certainly none of them are worth me holding up this PR for.  Even if it looks utterly obvious that [EdgeList](https://github.com/NetTopologySuite/NetTopologySuite/blob/0f56e478f91a0a40e4b75203ade55908990fc4ce/NetTopologySuite/GeometriesGraph/EdgeList.cs#L24), for example, **should** be OK with using a hash-based dictionary, it's not.  Because [OrientedCoordinateArray](https://github.com/NetTopologySuite/NetTopologySuite/blob/0f56e478f91a0a40e4b75203ade55908990fc4ce/NetTopologySuite/Noding/OrientedCoordinateArray.cs) does not have `Equals` / `GetHashCode` overrides that are consistent with its comparison behavior.

### DIFF
--- a/NetTopologySuite.IO/NetTopologySuite.IO.GeoTools/NetTopologySuite.IO.GeoTools.csproj
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.GeoTools/NetTopologySuite.IO.GeoTools.csproj
@@ -17,6 +17,7 @@
     <TargetFrameworkProfile>Client</TargetFrameworkProfile>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\</SolutionDir>
     <RestorePackages>true</RestorePackages>
+    <NeedsPowerCollections Condition=" '$(NeedsPowerCollections)' == '' ">False</NeedsPowerCollections>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
@@ -95,7 +96,7 @@
       <Project>{5770DAA9-84E5-4770-AF43-F6B815894368}</Project>
       <Name>NetTopologySuite</Name>
     </ProjectReference>
-    <ProjectReference Include="..\..\PowerCollections\Source\PowerCollections\PowerCollections.csproj">
+    <ProjectReference Condition=" '$(NeedsPowerCollections)' != 'False' " Include="..\..\PowerCollections\Source\PowerCollections\PowerCollections.csproj">
       <Project>{5821977d-ac2c-4912-bcd0-6e6b1a756167}</Project>
       <Name>PowerCollections</Name>
     </ProjectReference>
--- a/NetTopologySuite.Samples.Console/NetTopologySuite.Samples.Console.csproj
+++ b/NetTopologySuite.Samples.Console/NetTopologySuite.Samples.Console.csproj
@@ -45,13 +45,14 @@
     <TargetFrameworkProfile>Client</TargetFrameworkProfile>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
     <RestorePackages>true</RestorePackages>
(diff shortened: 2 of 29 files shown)


FObermaier:
+1

DGuidi:
merged

PR #85, opened 2016-02-13
Optimizations

netgrim:
Not sure why the GeoApi references where not there, 
but here's some changes that does basically the same in less lines


FObermaier:
GeoAPI project reference has been removed on purpose. The reference is achieved using its NuGet package and the `GeoAPI.Reference.targets` import at the top of the project file.
```
<Import Project="$(SolutionDir)BuildTools\GeoAPI.Reference.targets" Condition="Exists('$(SolutionDir)BuildTools\GeoAPI.Reference.targets')" />
```

Not sure about the other fixes you propose. Under normal conditions I'd suppose that the current implementation of the Length function would perform better than the one you propose, though your's is more readable.

PR #90, opened 2016-02-27
fix for nearest neighbor search for kd tree

lloydstockman:
from last commit "It was just checking to see if the left or right tree had a better score than current but the algorithm should check to see if the planes cross our best score for the coordinate we are looking for, not if the nodes score is better or not."

test added is from http://christopherstoll.org/assets/blogger/kD-tree_02.gif


DGuidi:
@FObermaier looks ok to me, any thoughts?It's something that can interest also @dr-jts?

PR #101, opened 2016-04-04
Fix for #100

### LINKED ISSUE #100: Too large buffer allocated in NetTopologySuite.IO.PostGisWriter
`NetTopologySuite.IO.PostGisWriter.GetByteStreamSize(IPolygon geometry, int coordinateSpace)` allocates much larger buffer whan needed for `geometry`, which may result even in `OutOfMemoryException` in some cases. The bug stems from iterating over IPolygon.InteriorRings, but taking the size of the whole collection in each iteration (instead of only current item), essentially leading to reporting size that is quadratic in regard to real size.

Yaevh:
fixes #100


DGuidi:
@FObermaier looks ok to me, van you take a look?

Issue #97, opened 2016-03-21
Defect in KdTree.NearestNeighbor


ikolev21:
Hello,

First of all, congratulations and best wishes for the great project.

I made some tests and found that `KdTree.NearestNeighbor` is not working correctly. I debugged the code, found the problem and fixed it, and also modified the algorithm a little to improve its performance.

Here's the unit test that I added to `KdTreeTest.cs`

        [Test]
        public void TestNearestNeighbor2()
        {
            var kd = new KdTree<string>();
            const int Count = 8;

            for (var row = 0; row < Count; ++row)
            {
                for (var column = 0; column < Count; ++column)
                {
                    kd.Insert(new Coordinate(column, row), (column * 100 + row).ToString());
                }
            }

            var testCoordinate = new Coordinate(Count / 2, Count / 2);
            var res = kd.NearestNeighbor(testCoordinate);

            Assert.AreEqual(testCoordinate, res.Coordinate);
        }

The returned result is 3,4 instead of the expected 4, 4.
I found that the smallest Count that exposes the problem is 8.
This was tested with the latest master branch.
I can provide a patch with my changes on request.

Best regards,
Ivan

DGuidi:
@ikolev21 can you create a [pull request](https://help.github.com/articles/using-pull-requests/) with your changes? So we can check modifies and compare with jts code.

ikolev21:
I don't have Git installed, I checked out the repository with TortoiseSvn. I don't want to install Git for now. So I suppose I cannot create a pull request. I can provide a patch.

For some reason attaching to this message wouldn't work, though I zipped the file as requested.
Here's a link to my Google drive:
https://drive.google.com/file/d/0B2MG4LKbRHikc215YWVHRXZtME0/view?usp=sharing

I couldn't figure out a better name for the test than TestNearestNeighbor2.

DGuidi:
thanks hope to take a look soon

DGuidi:
please can you check [revkdtree](https://github.com/NetTopologySuite/NetTopologySuite/compare/revkdtree) branch and validate my changes?
@FObermaier all tests are green, what do you think? Actually [i have no idea what i'm doing](https://goo.gl/rsLkGq)

DGuidi:
moved to #105

PR #105, opened 2016-04-14
KdTree patch

DGuidi:
Applied KdTree patch from #97

### DIFF
--- a/NetTopologySuite.TestRunner/Functions/SpatialIndexFunctions.cs
+++ b/NetTopologySuite.TestRunner/Functions/SpatialIndexFunctions.cs
@@ -27,7 +27,7 @@ namespace Open.Topology.TestRunner.Functions
 
         private static KdTree<object> BuildKdTree(IGeometry geom, double tolerance)
         {
-            var index = new KdTree<object>(tolerance);
+            var index = new KdTreeNTS<object>(tolerance);
             var pt = geom.Coordinates;
             for (int i = 0; i < pt.Length; i++)
             {
--- a/NetTopologySuite.TestRunner/Functions/SpatialIndexFunctions.cs
+++ b/NetTopologySuite.TestRunner/Functions/SpatialIndexFunctions.cs
@@ -27,7 +27,7 @@ namespace Open.Topology.TestRunner.Functions
 
         private static KdTree<object> BuildKdTree(IGeometry geom, double tolerance)
         {
-            var index = new KdTreeNTS<object>(tolerance);
+            var index = new KdTree<object>(tolerance);
             var pt = geom.Coordinates;
             for (int i = 0; i < pt.Length; i++)
             {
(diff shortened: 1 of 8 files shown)


DGuidi:
@ikolev21 feel free to add your feelings here

ikolev21:
Thanks, the patched stuff looks OK.

Now a couple of new comments :)
Look at line 196 that was commented out, `//if (currentNode != null) same check as in "while loop"`
I think this check was originally meant to be `if (_tolerance > 0)`.
Currently passing tolerance 0 to the tree doesn't work as expected because this check is missing, i.e. the tree will still merge points that have exactly the same coordinates, while the idea of tolerance 0 is to never merge points, even if coinciding.

The other comment is that the tolerance check can be made faster if the square of `_tolerance` is stored and comparison calculates just the squared distance instead of the exact one.

DGuidi:
please check also [JTS implemetation](https://github.com/dr-jts/jts/blob/f6187ee22aee29933c561284e1cb56df32fddcc0/modules/core/src/main/java/org/locationtech/jts/index/kdtree/KdTree.java).
Actually, I'm ambivalent about this changes: I'd like to have NTS strictly matches JTS code. So, maybe this code can be also proposed to @dr-jts for JTS integration?
I'm adding too much partecipants to this pullreq?

ikolev21:
I see that the JTS class misses the NearestNeighbor method, so our changes do not concern JTS.

As for `insertExact`, I think it has the same problem as NTS, the `if (currentNode != null) {` check must be changed to `if (tolerance > 0)`, to make tolerance==0 disable node merging.
The squared distance is less important, it's just a small performance optimization, but it still applies to JTS as well.

DGuidi:
@FObermaier sorry to involve you, but can you help me with this?
as said [here](https://github.com/NetTopologySuite/NetTopologySuite/pull/105#issuecomment-210288484) I'm ambivalent about this integration, any suggestion from you is welcome.

FObermaier:
As we have added this functionality in the first place, it should at least work alright.
Maybe we can refactor the code putting by putting the "non-JTS" portions in some partial class or make it an extension method.

IIRC I [submitted this enhancement to the mailing list](https://sourceforge.net/p/jts-topo-suite/mailman/message/34889739/). No response however.

DGuidi:
thx, missed submission to jts mailing list.

DGuidi:
I've made KdTree class abstract, with two different implementation for JTS (close to java code) and NTS (with fixes). If we like this changes, I'm ready to merge code.

FObermaier:
I don't like two implementations.
I checked, both splitting the source file and making `NearestNeighbor` an extension method to KdTree works. For the later we'll have to expose `_root` but that should not pose any probems.

DGuidi:
ah ok, made same changes. no prob.

PR #137, opened 2016-11-27
Refactoring/netstandard

acomarce:
If anyone has any use for this :), or parts of this

Notes:
- NetStander 1.6 only
- Updated syntax
- Needs lots more work


FObermaier:
Superceeded.

Issue #159, opened 2017-03-11
Newly created Shapefiles have invalid Last Update Date in header


NimbleKelly:
The LastUpdateDate property for DbaseFileHeader does not have a setter and no way to initialize it other than reading it from an existing file.  As a result, it is initialized with a default date of 01/01/0001.  When it is written out in the header, 1900 is subtracted from it.  This causes an invalid year, month, and day to be written and no way to specify any other value.

airbreather:
Can this be closed?

PR #160, opened 2017-03-11
Add public setter to LastUpdateDate property for DbaseFileHeader

NimbleKelly:
Newly-generated Shapefile headers have an incorrect last modified date since it always uses the default DateTime (1/1/0001) without a way to change it.  This addresses Issue [#159](https://github.com/NetTopologySuite/NetTopologySuite/issues/159).

### DIFF
--- a/NetTopologySuite.IO/NetTopologySuite.IO.GeoTools/Dbase/DbaseFileHeader.cs
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.GeoTools/Dbase/DbaseFileHeader.cs
@@ -59,12 +59,13 @@ namespace NetTopologySuite.IO
         }
 
         /// <summary>
-        /// Return the date this file was last updated.
+        /// Sets or returns the date this file was last updated.
         /// </summary>
         /// <returns></returns>
         public DateTime LastUpdateDate
         {
             get { return _updateDate; }
+            set { _updateDate = value; }
         }
 
         /// <summary>


FObermaier:
Is this the right aproach? Shouldn't it be set/updated in the `WriteShpHeader` function?

airbreather:
[WriteShpHeader](https://github.com/NetTopologySuite/NetTopologySuite/blob/dd40539808cc3a66073a3e23a9e70a4592baa356/NetTopologySuite.IO/NetTopologySuite.IO.GeoTools/ShapefileWriter.cs#L290-L302) doesn't look right to me... did you mean something else?

FObermaier:
@airbreather, you are right, `WriteShpHeader` has nothing to do with `LastUpdateDate` of `DbaseFileHeader`.

I wanted to say that `LastUpdateDate` should be updated prior to writing the `DbaseFileHeader`, instead of having a public property to set it.

airbreather:
I'm not sure it should be considered NTS's responsibility to set `LastUpdateDate` automatically, at least not without a way for the consumer to override that behavior.  In most use cases, I'm sure automatically setting it to something like `DateTime.Now` would be appreciated, but I imagine there are users who would like to at least have the option to set the value themselves.

For example: maybe the data is only up-to-date as of a week earlier than when the application is actually running, or maybe there's a time zone difference between where the software is being run and where the shapefile is expected to be consumed that causes it to be off by one.

PR #170, opened 2017-05-13
Fix #169

### LINKED ISSUE #169: .NET 2.0 and 3.5 Builds Are Broken
Steps to reproduce:
1. Clone latest `master` (c65a7ba at the time of writing).
2. `msbuild TeamCity.targets /t:BuildRelease20`

Expected: .NET 2.0 build succeeds.
Actual: failure, details look something like:

      Index\KdTree\KdTreeExtensions.cs(16,52): error CS1110: Cannot define a new extension method because the compiler required type 'System.Runtime.CompilerServices.ExtensionAttribute' cannot be found. Are you missing a reference to System.Core.dll?

Suggested fix: pull in [ExtensionAttribute.cs](https://github.com/dotnet/coreclr/blob/1413d462631f0d9bfc4c3df3cb5371e23bcbd8d9/src/mscorlib/shared/System/Runtime/CompilerServices/ExtensionAttribute.cs) and put it under Utilities/FrameworkReplacements/Runtime/CompilerServices

Doing that and rebuilding gets us another error:

    IO\Streams\ByteStreamProvider.cs(94,23): error CS1061: 'Stream' does not contain a definition for 'CopyTo' and no extension method 'CopyTo' accepting a first argument of type 'Stream' could be found (are you missing a using directive or an assembly reference?)

Suggested fix: semi-inline:

```cs
// 81920: largest multiple of 4096 that doesn't go on the LOH.
// 4096 is a common internal buffer size.
byte[] array = new byte[81920];
int count;
while ((count = input.Read(array, 0, array.Length)) != 0)
{
    ms.Write(array, 0, count);
}

return ms.ToArray();
```

However, after fixing that, both `BuildRelease20` and `BuildRelease35` fail to build NetTopologySuite.IO.GeoJSON with 112 errors.  A sample:

    Converters\AttributesTableConverter.cs(5,7): error CS0246: The type or namespace name 'Newtonsoft' could not be found (are you missing a using directive or an assembly reference?)

I do not have a suggested fix for this off the top of my head.

airbreather:
ExtensionAttribute: this wasn't added until .NET 3.5.
ByteStreamProvider: Stream.CopyTo wasn't added until .NET 4.0.
GeoJSON: reverting some changes from ee51f24, leaving only the version and HintPath updates.

### DIFF
--- a/NetTopologySuite.IO/NetTopologySuite.IO.GeoJSON/NetTopologySuite.IO.GeoJSON.csproj
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.GeoJSON/NetTopologySuite.IO.GeoJSON.csproj
@@ -53,8 +53,8 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Newtonsoft.Json.9.0.1\lib\net40\Newtonsoft.Json.dll</HintPath>
-      <Private>True</Private>
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.9.0.1\lib\$(JSON)\Newtonsoft.Json.dll</HintPath>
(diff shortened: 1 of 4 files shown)


DGuidi:
looks ok to me

Issue #26, opened 2014-12-04
Inspect PCL issues using .NET Portability Analyzer


DGuidi:
Microsoft released a new tool called [.NET Portability Analyzer](http://www.hanselman.com/blog/GettingReadyForTheFutureWithTheMicrosoftNETPortabilityAnalyzer.aspx) to help verify if written code is 100% portable.
I've briefly checked and some stuff, like Debug.WriteLine calls, are marked as errors.

PR #183, opened 2017-08-07
Resolves #110

### LINKED ISSUE #110: .NET Core 1.0 support
Anyone working towards supporting .NET Core?

airbreather:
Projects now require Visual Studio 2017 to build.
Many NTS projects now target .NET Standard 1.0 and 1.3.
NTS no longer supports .NET Framework 2.0 or any PCLs.
Clean up project files and add a properties file.
Handle multi-targeting with the new SDK format.
Conditional compilation is by feature, instead of by platform target.
Make APIs and behavior a bit more consistent across platform targets.
Remove some fringe APIs.

Other notes I took during my self-review to explain some of this stuff:

- GMLReader: more consistency between versions
- GMLWriter: more consistency between versions... please review this one specifically
- WKTFileReader: more consistency between versions; also, FileStream has had its own configurable buffer since .NET 1.1, so there's never been a need to wrap in BufferedStream
- ScaledNoder: very slight change... no longer checks the conditional in release mode because the inner block gets erased in release mode anyway
- <s>UnionInteracting: same as ScaledNoder</s> nope, there's a narrow scenario where this change could have caused NTS not to fail in places where JTS would fail.
- EdgeString: I get that JTS uses `ArrayList`, but it uses it exactly like `ArrayList<LineMergeDirectedEdge>` would look at runtime (because of Java's type erasure).  Besides, this should be slightly faster because of the lack of casting, and it's a lot more idiomatic.  I chose this over adding a package reference to `System.Collections.NonGeneric` for .NET Standard targets.
- SimpleNestedRingTester: see comments on EdgeString
- SweeplineNestedRingTester: see comments on EdgeString
- Caster: we don't need to use this anymore now that we dropped support for .NET 2.0 and PCL targets, but deleting this one is a bit awkward because it's public... with this change, old consumers that use these methods from old versions of NTS should continue to work with the newer version of NTS without needing to be recompiled, but things that use this will no longer compile after upgrading to the newer version of NTS.  Since .NET 2.0 and PCL consumers will no longer be able to use newer versions of NTS, everybody who would want to use these methods has a replacement that's not worse than what we have here (and usually better).  The error message will say exactly how to replace the obsolete methods.
- CollectionUtil: same basic comments as in Caster, but it probably needs a bit more justification because JTS has an equvialent to this.  As noted above, Java doesn't really have generic types, it just pretends to have them at compile-time.  So the non-generic stuff in JTS's CollectionUtil looks (for the most part) exactly like what it would look like to consumers.  This has not been idiomatic .NET for over a decade, and we do have idiomatic generic replacements for the JTS methods, so it seems appropriate to stop supporting those rather than keep code around that still uses ArrayList and adding a package reference to System.Collections.NonGeneric for .NET Standard targets.
- EnumUtility: note that I took the Enum.Parse call from the PCL side, because it's equivalent and more explicit.
- GeoToolsStreamTokenizer: this has been obsolete for over a decade since it was added in 2006.  Nobody should be using Geotools.NET together with NTS anyway, now that ProjNet4GeoAPI exists.
- EmbeddedResourceManager: no real reason to write to temp files all over the place when the PCL was doing the simpler thing all along: just use a stream
- Some of the automated tests will fail if we switch them to run against .NET Standard because the way NTS handles serialization won't work on all .NET Standard platforms.  For now, I've just removed it on those platforms; we will have to figure something out.

I'm also going to do more of this review myself, but it would annoy me way too much to leave #110 open without even having this pull request out there at this point.


airbreather:
As with [GeoAPI #26](https://github.com/NetTopologySuite/GeoAPI/pull/26), I haven't touched the travis stuff.

I also don't know how we will want to get the coded tests running for the other target platforms... there's a bit of a hint for making it work in the properties of NetTopologySuite.Tests.NUnit.csproj (note the separate `TargetFrameworkToTest`).

After this is done, I'm also thinking about moving some of those other side projects out of the main solution, as discussed [here](https://github.com/NetTopologySuite/NetTopologySuite/issues/110#issuecomment-308090406).  I sprung for taking down the target frameworks here instead of later, because that cuts down on how much has to get reviewed.

NetTopologySuite.Common.props
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <OutputPath>$(SolutionDir)$(Configuration)\$(Platform)</OutputPath>
+    <DocumentationFile>$(OutputPath)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <GeoAPIVersion>1.7.5</GeoAPIVersion>
```
airbreather: Better name?

NetTopologySuite.Common.props
```diff
@@ -0,0 +1,73 @@
...
+    <NETStandardPackageVersion>4.3.0</NETStandardPackageVersion>
```
airbreather: Better name?

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/ByteStreamProvider.cs
```diff
@@ -179,7 +191,11 @@ public void Dispose()
         private class ByteStream : MemoryStream
         {
             public ByteStream(ByteStreamProvider provider, bool writable)
+#if HAS_SYSTEM_IO_MEMORYSTREAM_CTOR_PUBLICLYVISIBLE
```
airbreather: Might as well lint these two spots while we're at it on L29/40...

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/FileStreamProvider.cs
```diff
@@ -19,10 +21,10 @@ public FileStreamProvider(string kind, string path, bool validatePath = false)
             if (path == null)
                 throw new ArgumentNullException("path");
 
-#if NET40
+#if HAS_SYSTEM_STRING_ISNULLORWHITESPACE
             if (string.IsNullOrWhiteSpace(path))
 #else
-            if (string.IsNullOrEmpty(path))
+            if (string.IsNullOrEmpty(path) || path.All(Char.IsWhiteSpace))
```
airbreather: Is there are more efficient helper method we can use instead of this LINQ crap, preferably without keeping the gap in functionality?

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/ShapefileStreamProviderRegistry.cs
```diff
@@ -56,10 +59,10 @@ public class ShapefileStreamProviderRegistry : IStreamProviderRegistry //, IDisp
         /// <param name="validateIndexPath">A value indicating that the shape index modified <paramref name="path"/> must be validated</param>
         public ShapefileStreamProviderRegistry(string path, bool validateShapePath = false, bool validateDataPath = false, bool validateIndexPath = false)
         {
-#if NET40
+#if HAS_SYSTEM_STRING_ISNULLORWHITESPACE
             if (string.IsNullOrWhiteSpace(path))
 #else
-            if (string.IsNullOrEmpty(path))
+            if (string.IsNullOrEmpty(path) || path.All(Char.IsWhiteSpace))
```
airbreather: See comment above, search "LINQ crap".

NetTopologySuite.Samples.Console/Operation/Polygonize/SplitPolygonExample.cs
```diff
@@ -115,7 +114,7 @@ internal static IGeometry ClipPolygon(IGeometry polygon, IPolygonal clipPolygona
 
 
         [STAThread]
-        public static void Main(string[] args)
+        public static void main(string[] args)
```
airbreather: I forget why I did this?  Probably a reason, but let's note it down at least...

NetTopologySuite.Samples.Console/Tests/Various/GraphBuilder.cs
```diff
@@ -1,137 +0,0 @@
-using System;
```
airbreather: Probably should have broken out "let's delete files that aren't actually in any of the projects" into their own step... oh well.

.gitignore
```diff
@@ -188,18 +188,15 @@ UpgradeLog*.htm
...
-NetTopologySuite.Samples.Shapefiles/__empty.shx
+/NetTopologySuite.Samples.Shapefiles/
 /NetTopologySuite/Algorithm/HCoordinate_DD.cs
 /NetTopologySuite/Algorithm/CGAlgorithmsAP.cs
 /NetTopologySuite.IO/NetTopologySuite.IO.GeoTools/NetTopologySuite.IO.GeoTools.csproj.DotSettings
-NetTopologySuite.Samples.Shapefiles/tmp*.shp
-NetTopologySuite.Samples.Shapefiles/tmp*.shx
-NetTopologySuite.Samples.Shapefiles/tmp*.dbf
-NetTopologySuite.Samples.Shapefiles/test_arcview.shp
-NetTopologySuite.Samples.Shapefiles/test_buffer.shp
```
airbreather: Surely, at the very least, **this line** was doing something for us?  As evidenced by the change to this file further below...

NetTopologySuite.TestRunner.Console/ConsoleTest.cs
```diff
@@ -146,6 +147,7 @@ static void RunOther()
         [STAThread]
         static void Main(string[] args)
         {
+            Environment.CurrentDirectory = new FileInfo(Assembly.GetEntryAssembly().Location).Directory.FullName;
```
airbreather: Comment, please.

NetTopologySuite.Tests.NUnit/IO/GML2/GMLReaderTest.cs
```diff
@@ -54,25 +55,51 @@ private static void DoTest(Type expectedType)
...
+                result.Add((reader, xml) =>
+                {
+                    XmlDocument doc = new XmlDocument();
+                    doc.LoadXml(xml);
+                    return (IGeometry)xmlDocMethod.Invoke(reader, new object[] { doc });
+                });
+            }
+
+            MethodInfo xDocMethod = typeof(GMLReader).GetMethod("Read", new Type[] { typeof(XDocument) });
+            if (xDocMethod != null)
```
airbreather: I think after getting rid of `NET20` and `PCL`, this is always non-null, but I guess it doesn't hurt anybody.

NetTopologySuite.Tests.NUnit/Performance/Geometries/Prepared/PreparedPolygonIntersectsPerformanceTest.cs
```diff
@@ -129,7 +129,7 @@ private static void Test(IGeometry g, ICollection<IGeometry> lines)
                 var count2 = TestPrepGeomCached(g, lines);
                 sw.Stop();
                 time2 += sw.ElapsedMilliseconds;
-
+
```
airbreather: Pointless.  Revert.

NetTopologySuite.Windows.Forms.Test/BasicTest.cs
```diff
@@ -1,99 +1,99 @@
...
-                g.FillPath(new SolidBrush(Color.FromArgb(180, Color.Red)), res);
-                g.DrawPath(Pens.Red, res);
-            }
-            b.Save(string.Format("MultiPoint-{0}.png", factory), ImageFormat.Png);
-            
-
-        }
-    
-    }
+using System;
```
airbreather: Something weird happened that marked this whole file as changed on GitHub... why?  Was it an encoding thing?  Line endings?  Not looking at it until it's resubmitted without every line being marked as deleted + added, or minimally an explanation why.

NetTopologySuite.Windows.Forms.Test/BasicTestMedia.cs
```diff
@@ -1,374 +1,374 @@
...
-            PngBitmapEncoder png = new PngBitmapEncoder();
-            png.Frames.Add(BitmapFrame.Create(rtb));
-            using (Stream stm = File.Create(fileName))
-            {
-                png.Save(stm);
-            }
-        }
-    }
-
+﻿using System;
```
airbreather: Something weird happened that marked this whole file as changed on GitHub... why?  Was it an encoding thing?  Line endings?  Not looking at it until it's resubmitted without every line being marked as deleted + added, or minimally an explanation why.

NetTopologySuite/Geometries/Implementation/DotSpatialAffineCoordinateSequence.cs
```diff
@@ -20,8 +20,8 @@ public class DotSpatialAffineCoordinateSequence :
         private readonly double[] _m;
         
         private readonly Ordinates _ordinates;
-        
-#if !PCL
+
+#if HAS_SYSTEM_SERIALIZABLEATTRIBUTE
```
airbreather: Is it right to use the same constant in both places?

NetTopologySuite/IO/GML2/GMLWriter.cs
```diff
@@ -84,7 +71,7 @@ public void Write(IGeometry geometry, Stream stream)
         /// </summary>
         /// <param name="coordinate"></param>
         /// <param name="writer"></param>        
-        protected void Write(Coordinate coordinate, XmlTextWriter writer)
+        protected void Write(Coordinate coordinate, XmlWriter writer)
```
airbreather: Will this change work if there are any subclasses compiled against older versions of the library that call the older version of the method?  If not, then this is technically a breaking change, but is that a big deal?

NetTopologySuite/Index/KdTree/KdTreeExtensions.cs
```diff
@@ -77,8 +77,11 @@ public static class KdTreeExtensions
             if (isOddLevel)
                 return (left ? target.X <= node.X : target.X >= node.X) || Math.Pow(target.X - node.X, 2) < closestDistSq;
             return (left ? target.Y <= node.Y : target.Y >= node.Y) || Math.Pow(target.Y - node.Y, 2) < closestDistSq;
+
+/*
```
airbreather: Delete this instead?  What does it look like in JTS?

NetTopologySuite/Operation/Union/UnionInteracting.cs
```diff
@@ -44,7 +44,7 @@ public IGeometry Union()
             // check for all interacting or none interacting!
             IGeometry int0 = ExtractElements(_g0, _interacts0, true);
             IGeometry int1 = ExtractElements(_g1, _interacts1, true);
-#if !PCL
+#if DEBUG
```
airbreather: Can `IsEmpty` throw, and/or can either `int0` or `int1` be `null`?  If the answer to either is true, then this is technically a breaking change; if so, does that matter?

NetTopologySuite/Utilities/BitTweaks.cs
```diff
@@ -48,7 +42,7 @@ internal static float ReverseByteOrder(float value)
 
         internal static double ReverseByteOrder(double value)
         {
-            return DoubleBitsConverter.Int64BitsToDouble(ReverseByteOrder(DoubleBitsConverter.DoubleToInt64Bits(value)));
+            return System.BitConverter.Int64BitsToDouble(ReverseByteOrder(System.BitConverter.DoubleToInt64Bits(value)));
```
airbreather: Why fully-qualify here?

NetTopologySuite.Common.props
airbreather: Going with `GeoAPIPackageReferenceVersion`

NetTopologySuite.Common.props
airbreather: Going with `NETStandardPackageReferenceVersion`

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/FileStreamProvider.cs
airbreather: `StringEx.IsNullOrWhiteSpaceNTS` uses a regex, but this should not be awful: in almost all cases, I expect it to terminate after reading exactly one character, and this all happens very close to filesystem operations that drown out this cost, so I'm going to leave it alone instead of either try to reference that method (it's internal in a different assembly) or copy the regex here.

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/ShapefileStreamProviderRegistry.cs
airbreather: Same reply as in the comment above.

NetTopologySuite.Samples.Console/Operation/Polygonize/SplitPolygonExample.cs
airbreather: Most likely, it was because I hadn't done [this](https://github.com/NetTopologySuite/NetTopologySuite/pull/183/files#diff-d6393584118e2f28680d7a7d1c9e98f6R8) yet (copied over the `StartupObject` from the project's properties).

Reverting.

.gitignore
airbreather: Looks like re-running `TestReadingShapeFileAfvalbakken` on its own, the file gets reverted back to what it was before the change, so it's probably ideal to keep it **out** of `.gitignore`.

NetTopologySuite.Windows.Forms.Test/BasicTest.cs
airbreather: Downloading the raw files from GitHub, GitHub is serving LF line endings for the old file and CRLF line endings for the new one.

NetTopologySuite/Geometries/Implementation/DotSpatialAffineCoordinateSequence.cs
airbreather: I think so.

NetTopologySuite/IO/GML2/GMLWriter.cs
airbreather: Open question for other reviewers.

NetTopologySuite/Index/KdTree/KdTreeExtensions.cs
airbreather: I can't find an equivalent in JTS.  Open question for other reviewers.

NetTopologySuite/Operation/Union/UnionInteracting.cs
airbreather: I'm not interested in defending this change.  JTS calls `isEmpty()` no matter what.  NTS can do the same.

NetTopologySuite/Utilities/BitTweaks.cs
airbreather: Because `NetTopologySuite.Utilities.BitConverter` exists.  Obsoleting that one now.

airbreather:
@xivk @DGuidi @FObermaier 

Please review this PR (not sure who makes the most sense, or if everybody should chip in).  It looks a bit big, but the majority of it is just deleting cookie-cutter stuff from .csproj files and replacing it with different, smaller, human-readable cookie-cutter stuff.

If absolutely nothing else, I'd at least like some input on these open questions from my own review:

https://github.com/NetTopologySuite/NetTopologySuite/pull/183#discussion_r131934372
https://github.com/NetTopologySuite/NetTopologySuite/pull/183#discussion_r131935161

DGuidi:
Hope to take a look as soon as possible, not sure I can help actually

NetTopologySuite/IO/GML2/GMLWriter.cs
DGuidi: subclasses that override method using "new" should be broken, not a big deal.

DGuidi:
kudos to all of you, maybe now I undestood how a true PR is made. thanks again

NetTopologySuite/Index/KdTree/KdTreeExtensions.cs
DGuidi: unable to find a reference in jts: maybe old code now removed?

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/FileStreamProvider.cs
DGuidi: >  I expect it to terminate after reading exactly one character

so the same is valid for LINQ. personally, I'd like this linq functions, but feel free to use what do you like.

NetTopologySuite.Samples.Console/Tests/Various/GraphBuilder.cs
DGuidi: this is more an experiment, maybe we can clean this "crap" now...

.gitignore
DGuidi: there are test artifacts, afaik.

NetTopologySuite/Utilities/BitTweaks.cs
DGuidi: > NetTopologySuite.Utilities.BitConverter
probably because PCL (or Compact Framework) doesn't have SystemBitConverter?

DGuidi:
travis build is broken, maybe we can fix travis build later?

DGuidi:
I wanna wait for @FObermaier and @xivk suggestions, before integrate this stuff into master...
@pauldendulk maybe can do a review, too?

Side note: shouldn't this stuff go in develop branch?

airbreather:
> as a side note: shouldn't this stuff go in develop branch, instead of master?

Didn't realize that that had happened yet.  I just switched the base branch so it targets `develop` now.

xivk:
This doesn't build, perhaps I'm going it wrong but I would expect opening it in VS and building it should work.

@airbreather Can you elaborate on this:

<PackageReference Include="GeoAPI" Version="$(GeoAPIPackageReferenceVersion)" />

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/FileStreamProvider.cs
xivk: I'm assuming this one is not available:

https://msdn.microsoft.com/en-us/library/system.string.isnullorwhitespace(v=vs.110).aspx

LINQ is usually bad for performance but because this is checking a path I'm assuming this happens only once when opening a filestream and won't have that much impact any.

I would just leave this in.

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/ShapefileStreamProviderRegistry.cs
xivk: Same as above.

NetTopologySuite/IO/GML2/GMLWriter.cs
xivk: I think this is fine...

NetTopologySuite/Index/KdTree/KdTreeExtensions.cs
xivk: If it was there in the old version then leave it I would say.

We can cleanup code later, this is just about getting .NET Core support.

.gitignore
xivk: It doesn't hurt anyone to have these lines in and tests may be stopped in the middle of them running.

NetTopologySuite.Samples.Console/Tests/Various/GraphBuilder.cs
xivk: We can do cleanup of the code later, let's not try to achieve absolute perfection on this pullrequest alone... ;-)

xivk:
Ok, so I checked a bit more dotnet restore works, but packages don't restore in VS.

Anyway, it should just work when opening the project.

NetTopologySuite/Index/KdTree/KdTreeExtensions.cs
DGuidi: > We can cleanup code later
yep, I agree

DGuidi:
there's a simpler way to check the build in my PC, other to clone [squashed](https://github.com/airbreather/NetTopologySuite/tree/squashed) repository?

DGuidi:
checked: same build errors as @xivk

DGuidi:
teamcity.targets should be deprecated/removed, i think

DGuidi:
@xivk GeoAPIPackageReferenceVersion is defined in NetTopologySuite.Common.Props

NetTopologySuite.Samples.Console/Tests/Various/GraphBuilder.cs
airbreather: Actually, files like this one had to be handled somehow, because the new SDK projects automatically include all files in the folder structure, whereas the old project files forced you to write out every single file that you wanted included.  This particular file (along with almost all of the other files that I outright deleted) was subtly omitted from the old monster .csproj files, so it just sat there doing nothing, not working.

So if I didn't delete this (and others), then I would instead have to do something like
```xml
<Compile Remove="Path\To\ThisFile.cs" />
```

Which seemed kinda silly.

NetTopologySuite.IO/NetTopologySuite.IO/IO/Streams/FileStreamProvider.cs
airbreather: Yeah, it also only happens in the .NET 3.5 target, so I expect that almost nobody will hit this block anyway.

airbreather:
> This doesn't build, perhaps I'm going it wrong but I would expect opening it in VS and building it should work.

> checked: same build errors as @xivk

In an [earlier comment](https://github.com/NetTopologySuite/NetTopologySuite/pull/183#issuecomment-320810731), I'd mentioned:
> Finally, it assumes that exactly GeoAPI 1.7.5 is in a local NuGet package source, which doesn't exist yet (it expects the absolute latest of GeoAPI)... the package version is in NetTopologySuite.Common.props, just in that one place, so if you want to push out a newer pre than what was out there when I started this, feel free to do so and then change that one spot.

Is that the cause of the build errors?

To elaborate, on my local machine, I have a locally-built package in a local NuGet source that satisfies the GeoAPI 1.7.5 package reference.  NuGet doesn't appear to have anything compatible, as it requires the stuff that @FObermaier merged in NetTopologySuite/GeoAPI@273e00f.

With that package in my local source, a fresh clone of the `squashed` branch is able to build in Visual Studio 2017 Community 15.3 without any other changes.  If someone wanted to push a new GeoAPI, maybe `1.7.5-pre3`, to NuGet, then I can push a package reference version change.

> teamcity.targets should be deprecated/removed, i think

I agree... I just left it alone because I'm not sure how the build environment is all set up.  It definitely will need to be at least somewhat updated if nothing else.

.gitignore
airbreather: It looks like I just copied all of `.gitignore` from 94d51b3.  I think I should just revert the changes for this file, entirely.

airbreather: 5830f44 reverts all changes to `.gitignore`.

NetTopologySuite/Index/KdTree/KdTreeExtensions.cs
airbreather: Reverted this.  Looks like I'd taken it from [here](https://github.com/NetTopologySuite/NetTopologySuite/commit/76cc663f430b9323475c7592d7b79d0cb91930f6#diff-0df2957c80cc061d734fc162aa7a2781), and I approved of this particular change (commenting out provably dead code) even though this file no longer otherwise **needs** to be changed.

NetTopologySuite/Utilities/BitTweaks.cs
airbreather: I think PCL didn't have the two methods that `System.BitConverter` has because Silverlight 5 didn't let you do unsafe code, which those methods rely on... now that we don't have to target PCL, we **could** outright delete `NetTopologySuite.Utilities.BitConverter`, but that would be a bit of a breaking change since the latter is public.  The `[Obsolete(..., error: true)]` approach I took should make the upgraded NTS compatible with old binaries that use our `BitConverter` for some reason, while still requiring dependents to switch to using `System.BitConverter` when building from source.

xivk:
I was working until now on GeoAPI, setting up proper build and packages publishing, I assumed it was already ok with this package for now:

https://www.nuget.org/packages/GeoAPI/1.7.5-pre2

I updated my VS install to 15.3 and pulled in the latest changes and things seem to work now. Still testing.

xivk:
If I use 1.7.5-pre2 things work fine, for me the pullrequest is fine, it's going into develop anyway and a release has to wait until we have a 1.7.5 for GeoAPI anyway.

xivk:
So do we more or less agree we can merge then I can get started on the builds and packages? 

The issue seems to be an unupdated instances of VS 2017, I'm also going to add build scripts for those who have just .NET Core and VS Core installed, adding linux/macos support once merged. Prereleases for now but we could have stables pretty soon if we want.

If there are no more responses within a few hours I'm merging. It's merging in the dev-branch anyway for now so the risk is low.

airbreather:
I was hoping to wait for @FObermaier to at least give some input on it.

If you're all OK with where it is right now and you don't think it's necessary to wait for @FObermaier to look at it, then I'm fine with doing one of those GitHub "Squash and Merge" things (I'd really like a squash here because I **think** that it should unmark certain files as being changed, which would make the history look more relevant).

DGuidi:
I'm with @airbreather and I will wait for some input from felix

DGuidi:
> If someone wanted to push a new GeoAPI, maybe 1.7.5-pre3, to NuGet, then I can push a package reference version change.

why not using pre2? @xivk said that all looks ok with this package.

> on my local machine, I have a locally-built package in a local NuGet source that satisfies the GeoAPI 1.7.5 package reference

sorry, can you share how to do this? or even better a zip to download and unzip? I'm unable to do this alone (yep, you're right: i'm stick with plain old "legacy" .net and newer stuff are obscure to me)

airbreather:
@DGuidi  I'm at work now so I'll respond to your messages when I get home.  I just wanted to respond to @xivk's time-based message quickly.

DGuidi:
@airbreather feel free to answer when you've time, of course. I'm in vacation actually, so I'm checking now just because I'm bored :)

airbreather:
> why not using pre2? @xivk said that all looks ok with this package.
I guess it's fine.  I hadn't tried it out because I'd figured that there'd be **some** kind of problem that I would run into, but the worst I'm seeing with `1.7.5-pre2` is that when you switch the platform target you're looking at in VS2017 to one of the `netstandard1.x` targets, you lose useful XML docs in the IntelliSense for GeoAPI types, because for some reason `1.7.5-pre2` didn't ship GeoAPI.xml.

> some unexpected tests fail (like `OctagonalEnvelopeTest.TestRobust`)

That test fails for me on a fresh clone of `develop` as well, on my machine.

>> on my local machine, I have a locally-built package in a local NuGet source that satisfies the GeoAPI 1.7.5 package reference
> sorry, can you share how to do this? or even better a zip to download and unzip? I'm unable to do this alone (yep, you're right: i'm stick with plain old "legacy" .net and newer stuff are obscure to me)

I did it like this:
1. build GeoAPI.sln in Visual Studio, release mode
    1. or from a Developer Command Prompt for VS 2017 window, you could just run `msbuild GeoAPI.sln /t:Restore;Rebuild /p:Configuration=Release`
2. using a standalone version of NuGet (my version is 4.1.0.2450), run `nuget pack GeoAPI.nuspec -Version 1.7.5 -Symbols`
3. put the resulting GeoAPI.1.7.5.nupkg file into one of the folders listed in Visual Studio's `Tools > Options > NuGet Package Manager > Package Sources` (you might need to add a folder to this list if you haven't added any before... there's a "+" button there).

DGuidi:
actually `OctagonalEnvelopeTest.TestRobust` should fail, I've added this test some time ago as part of JTS alignment and this test is a failure case, so... my bad :(
I'm adding some attributes for tests that should fail

DGuidi:
reviewed some code in develop branch, maybe squashed should be realigned

DGuidi:
note that tests in `NetTopologySuite.IO.Streams.CloudStorage.Test` are ignored in develop branch (see TestFixtureSetup)... I see failures in squashed branch instead

DGuidi:
A lot less tests in squashed branch, I assume related to 

> Some of the automated tests will fail if we switch them to run against .NET Standard because the way NTS handles serialization won't work on all .NET Standard platforms. For now, I've just removed it on those platforms; we will have to figure something out.

to me, we can go ahead, but we keep track of this and reintegrate soon

airbreather:
> A lot less tests in squashed branch, I assume related to

>> Some of the automated tests will fail if we switch them to run against .NET Standard because the way NTS handles serialization won't work on all .NET Standard platforms. For now, I've just removed it on those platforms; we will have to figure something out.

> to me, we can go ahead, but we keep track of this and reintegrate soon

I did a side-by-side comparison before.  I'll look into `NetTopologySuite.IO.Streams.CloudStorage.Test` after making this comment (I'm seeing the same thing you're seeing now), but yes, the main reason for the difference in the number of tests is related to the quoted comment.

Previously, there was this special `NetTopologySuite.Tests.NUnit.Pcl` project that ran against the special PCL projects, with mostly the same exact test files.  When switching like this, I had set it all up in e58b42e so that instead of just tests for the two targets of .NET 4.0 and PCL, there were tests for all the different targets (it was just a few lines of .csproj for each different target).

The main problem with all that was that with all the different test projects for the different target frameworks, all the tests took **a lot** longer to run, and the majority of the tests didn't have any platform-specific differences so it would probably be a waste to just blast most of the tests 6x each.

# Squashed
as of 65c0cefb
image

# Develop
as of 857e4c17
image

airbreather:
> note that tests in NetTopologySuite.IO.Streams.CloudStorage.Test are ignored in develop branch (see TestFixtureSetup)... I see failures in squashed branch instead

Ha.  OK.  Got it.

In `develop`, there is a mismatch between the `packages.config` files.

[non-test](https://github.com/NetTopologySuite/NetTopologySuite/blob/88005f62cfe61a130e1cca30b73beea04efa32c5/NetTopologySuite.IO/NetTopologySuite.IO.Streams.CloudStorage/packages.config#L9): 5.0.0
[test](https://github.com/NetTopologySuite/NetTopologySuite/blob/e36958b4d1f67311312e957393241f650277b0b9/NetTopologySuite.IO/NetTopologySuite.IO.Streams.CloudStorage.Test/packages.config#L11): 5.0.2

If everything was done by-the-book, then the tests would be failing in `develop`.

However, the non-test version is sneaky.  Its own `packages.config` file restores 5.0.0, but then it abuses the fact that the test's `packages.config` file causes us to restore 5.0.2, and has its actual assembly reference use that restored version.

When moving to the `PackageReference`, I just transferred the data from `packages.config`, but didn't notice this mismatch.  These tests were all a sort of "broken" before, so I paid no attention to them.

Anyway, excuses aside, these tests are back to "ignored" after 40115d5.

xivk:
Ok than I guess we are just waiting for feedback from @FObermaier 

I think we pretty much have all this working in the same way as before, except refactored to the new csproj format support .NET Core. That's what it comes down to right?

DGuidi:
@airbreather I've do this:
1. manage nuget for solution
2. 'consolidate' azure.storage

some changes in a lot of csproj, then tests ignored correctly

airbreather:
> I think we pretty much have all this working in the same way as before, except refactored to the new csproj format support .NET Core. That's what it comes down to right?

Sorta (well, .NET **Standard** so all the platforms from [this grid](https://github.com/dotnet/standard/blob/643967c317668b4dd8450897aa6401b42b14359d/docs/versions.md), but same idea).  There were also a few other changes as mentioned in the original pull request comment, biggest ones being to remove `NET20` and `PCL` targets and to deprecate things that were only needed by those targets.

It also makes the typical out-of-the-box VS2017 "Build Solution" build for all target platforms at once, instead of just exactly `NET40` + `PCL` and then requiring a manual step to build the other target platforms, since it would have been harder **not** to do that than to try to keep that part of this process around.

DGuidi:
thanks for all the clarifications and support.

airbreather:
Ahh!  I've figured out why I really wanted to use the newer GeoAPI package.  `1.7.5-pre2` references the NETStandard.Library metapackage, which I really didn't want to do for NTS recursively.

I touched on this a bit earlier, but although referencing NETStandard.Library is the current recommendation for new .NET Standard libraries, and it's the only way to do a .NET Standard 2.0 library, it can cause frustration when trying to reference a mixed bag of .NET Standard / .NET Framework packages, at least if you're not using the latest SDK tooling.

So now that I have access, I'm going to push new pre-release GeoAPI and NTS packages that are all @airbreather approved to minimize frustration, the best way I know how...

airbreather:
Hmm.  Re-running **all** the tests one last time (I've been mainly focusing on the stuff in `NetTopologySuite.Tests.NUnit`), I noticed that the stuff in `Issue47Tests` is failing in my branch, and not in the base.

I'd changed GMLWriter to look more like the PCL version instead of the .NET version, so I looked to confirm that vanilla PCL fails the same way on the same inputs... and it does.

airbreather:
All right.  All tests pass, except those that are explicitly indicated in the source code for those tests as things I shouldn't expect to see passing.

I'm going to just squash-merge this into `develop`, push pre-release packages onto NuGet for the packages that now have proper .NET Standard support, and then start looking into what it takes to get Travis doing the right thing now that @xivk seems to have gotten it working nicely for GeoAPI.  Like @xivk said, pushing to `develop` is fairly low-risk, and it's all pre-release stuff.

If @FObermaier does his own review and finds issues, then we can always just make a few more commits to fix them.  It's not like it's all that likely that he's going to find something that makes us have to start from scratch on this, anyway.

plamen-i:
@airbreather  Maybe I'm missing something, but still a question/note - do you intentionally did not upgrade Newtonsoft.Json to latest version (in NetTopologySuite.IO.GeoJSON)? And maybe remove upper limit on its version (>= instead of =)?

airbreather:
> @airbreather Maybe I'm missing something, but still a question/note - do you intentionally did not upgrade Newtonsoft.Json to latest version (in NetTopologySuite.IO.GeoJSON)? And maybe remove upper limit on its version (>= instead of =)?

I didn't want to make that kind of a change here.  There were already too many other seemingly "unnecessary" changes that were hard to defend, without upgrading adding in bumping up the version number on a critical dependency.

DGuidi:
> do you intentionally did not upgrade Newtonsoft.Json to latest version (in NetTopologySuite.IO.GeoJSON)? And maybe remove upper limit on its version (>= instead of =)?

this and other similar revisions probably should be done later

plamen-i:
My intention was just to remind of this because when mixing with other packages in the same project this can be a problem. I'm OK for now but I will be happy if this is done soon after finalizing current changes:-)

PR #193, opened 2017-09-29
Fix #192

### LINKED ISSUE #192: NullReferenceException in PackedCoordinateSequence.GetCachedCoords
Just got an exception with this at the top of the stack trace:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at NetTopologySuite.Geometries.Implementation.PackedCoordinateSequence.GetCachedCoords()
   at NetTopologySuite.Geometries.Implementation.PackedCoordinateSequence.ToCoordinateArray()
   at NetTopologySuite.Algorithm.Centroid.Add(IPolygon poly)
   at NetTopologySuite.Algorithm.Centroid.Add(IGeometry geom)
   at NetTopologySuite.Geometries.Geometry.get_Centroid()
```

Looks like [this stuff](https://github.com/NetTopologySuite/NetTopologySuite/blob/65f5206a002e62b3e4ec173d6ab197e581c2fa99/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs#L147-L156) should fetch `CoordRef` into a local before null-checking, since this looks like a race condition otherwise.

airbreather:
@dr-jts I think this applies equally well to JTS, around [these lines](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java#L104-L116).

The problem is an apparent race condition: `CoordRef` (`coordRef` in JTS) can be modified between the null-check and when we actually fetch the reference target.  It seemed to happen not-unreliably when I was doing stuff at work today, though I didn't get the opportunity to confirm that this exact change fixed it in the app (though, what else in this method could possibly throw `NullReferenceException`?).

This could happen even if nobody's actually mutating the geometry: threads A and B could both just be calling `GetCachedCoords` (`getCachedCoords` in JTS), they could both run just past the initial null check, then Thread A could go all the rest of the way through the method and null out the field, then when it's Thread B's turn, it re-fetches the field (which is now null) and then tries to dereference.

If not for the fact that JTS seems to have the same problem, I'd just merge it in myself, but as-is, I'd like to get some feedback from others because of how touchy we are about deviating from JTS...

### DIFF
--- a/NetTopologySuite/Geometries/Implementation/DotSpatialAffineCoordinateSequence.cs
+++ b/NetTopologySuite/Geometries/Implementation/DotSpatialAffineCoordinateSequence.cs
@@ -274,9 +274,10 @@ namespace NetTopologySuite.Geometries.Implementation
         /// <returns></returns>
         private Coordinate[] GetCachedCoords()
         {
-            if (_coordinateArrayRef != null && _coordinateArrayRef.IsAlive)
+            var localCoordinateArrayRef = _coordinateArrayRef;
+            if (localCoordinateArrayRef != null && localCoordinateArrayRef.IsAlive)
             {
-                var arr = (Coordinate[])_coordinateArrayRef.Target;
+                var arr = (Coordinate[])localCoordinateArrayRef.Target;
                 if (arr != null)
                     return arr;
 
--- a/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
+++ b/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
@@ -144,11 +144,16 @@ namespace NetTopologySuite.Geometries.Implementation
         /// <returns></returns>
         private Coordinate[] GetCachedCoords()
         {
-            var arr = (Coordinate[]) CoordRef?.Target;
-            if (arr != null) 
-                return arr;
+            var localCoordRef = CoordRef;
+            if (localCoordRef != null)
+            {
+                var arr = (Coordinate[]) localCoordRef.Target;
+                if (arr != null)
+                    return arr;
 
-            CoordRef = null;
+                CoordRef = null;
+                return null;
+            }
             return null;
         }
 
--- a/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
+++ b/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
@@ -144,15 +144,11 @@ namespace NetTopologySuite.Geometries.Implementation
         /// <returns></returns>
         private Coordinate[] GetCachedCoords()
         {
-            if (CoordRef != null) 
-            {
-                var arr = (Coordinate[]) CoordRef.Target;
-                if (arr != null) 
-                    return arr;
-                
-                CoordRef = null;
-                return null;
-            }
+            var arr = (Coordinate[]) CoordRef?.Target;
+            if (arr != null) 
+                return arr;
+
+            CoordRef = null;
             return null;
         }
 


dr-jts:
Well, not surprising, cached refs are tricky to get right under concurrency.

Java doesn't have the null-conditional operator.  So perhaps synchronizing part or all of the method is the best equivalent?  Does C# have the synchronized declaration?  if not, this might be a point of divergence.

airbreather:
An equivalent would just be to load the field value into a local variable first and then do both the null-check and the `.get()` on that local variable reference.

The null-conditional operator here is just syntactic sugar that compiles to exactly the same thing.

dr-jts:
Reading [this SO post](https://stackoverflow.com/questions/27628273/test-a-weak-reference-before-using-it-java) made me think that perhaps the real problem is that coordRef should _never_ be nulled out, but only refreshed if its ref turns out to be null.  Then the pattern from the post can be used, and there is less complex logic about state.  Thoughts?

airbreather:
Well, it's necessary to do *something* with `coordRef` in `setX` / `setY` / `setOrdinate`, for [correctness reasons](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java#L30-L31).  I'm not sure I like the idea of making every setter method call in PackedCoordinateSequence to have to go through up to 3 pointer dereferences to update the underlying Coordinate instance instead of just nulling out the ref and letting it get regenerated when Coordinate instances are actually needed, but it's obviously your call.

I agree that `coordRef` doesn't *need* to [get nulled out in `getCachedCoords`](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java#L110) in order to be correct, though I figured that a possible reason to do so was to slightly reduce memory usage when [`toCoordinateArray()`](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java#L95) is called occasionally but infrequently, and [`getCoordinate(int i)`](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java#L60) is called frequently.

dr-jts:
Those are all good observations!  I agree that it's better to be lazy about initializing the cache, and setting it to `null` is an obvious way to do that. (It would be nicer if this behaviour was encapsulated in its own class, but that would introduce another object.)

As for the issue of unexpected mutation, the CoordinateSequence API makes no promises about the effect of mutation on aliased objects.  So this issue should not carry much weight in the design.  Perhaps an immutable CoordinateSequence concept would be useful, but I think it's best to **not** encourage relying on subtle nuances of implementation behaviour. 

So your suggestion of capturing the `coordRef `reference is still the best one.

FObermaier:
@airbreather I support merging this. You have communicated the issue you are having and the fix is approved valid.

dr-jts:
@airbreather Any thought on the option of using `synchronized` (`lock` in C# I think)?  One potential advantage is that this prevents repeating the generation of the cache when two threads are racing each other.

dr-jts:
@airbreather I'm curious about what in your code is triggering this race failure?  A goal in JTS is to eliminate the use of `getCoordinates()` in favour of direct access to the ordinates (which is more compatible with the CoordinateSequence paradigm).  It's useful to know if there are hotspots in the library in order to prioritize rewriting calls getCoordinates().

airbreather:
> Any thought on the option of using synchronized (lock in C# I think)? One potential advantage is that this prevents repeating the generation of the cache when two threads are racing each other.

As an *opinion*, I'm inclined to vote against that, for these reasons:

1. There's such a narrow window of time where this can even cause a failure.
2. Given how narrow this window is, the cost of duplicate rebuilds that internal synchronization would have prevented feels like it would amortize to essentially 0, especially given the next two.
3. This seems very likely to find itself on the "hot path" for many applications that only do accesses from a single thread.  Because (*I think*) synchronization implies a memory barrier, this feels like it could **really** hurt the common case.
4. JTS's version uses a `SoftReference`, which is much less likely than NTS's `WeakReference` to be freed, so NTS is **much** more likely than JTS to see this problem (as evidenced by how long JTS has been around without ever seeing this).
5. Parallel application threads where some mutate the underlying sequence and others check `getCoordinates()` / etc. are not going to be correct without external synchronization anyway (minimally, we might be in the middle of a "write X, then write Y" series, which no internal synchronization can defend against).
6. Per the goal of JTS to use the `CoordinateSequence` directly, instead of `Coordinate` instances, I would expect the already narrow list of cases where this can happen to get even narrower in future releases.

> I'm curious about what in your code is triggering this race failure? 

The code that's actually trying to dereference a null would be the NTS equivalent to [`Centroid.add(Polygon)`](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/algorithm/Centroid.java#L159-L165).  I'm not immediately sure what else is coincidentally reading the coordinates at the same time.

I've seen the issue twice in a row, the only two times I ran with that particular set of inputs, which is enough for me.  Given that it takes several minutes to get to the repro in the app I'm running with, and that the problem did go away after applying this, I'm not enthusiastic about trying to hunt down any other parallel threads that might coincidentally be trying to fetch the cache.

Like I mentioned above, though, JTS is far less susceptible to this than NTS is, because of the difference between Java's `SoftReference` and the CLR's `WeakReference` (which is the same as Java's `WeakReference`).

> A goal in JTS is to eliminate the use of `getCoordinates()` in favour of direct access to the ordinates (which is more compatible with the CoordinateSequence paradigm). It's useful to know if there are hotspots in the library in order to prioritize rewriting calls getCoordinates().

You may be interested in 6a7ce8f, which modifies the equivalents to [these](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/algorithm/CGAlgorithms.java#L109-L148) two [spots](https://github.com/locationtech/jts/blob/e570b7621cb629639d53a47822a67120e9fa7a90/modules/core/src/main/java/org/locationtech/jts/algorithm/PointLocator.java#L182) in JTS.  Since the CLR doesn't do escape analysis like the JVM does, this was making `PointLocator`, and thus `OverlayOp`, disproportionately awful in NTS.

dr-jts:
> As an opinion, I'm inclined to vote against that, for these reasons:

All very good reasons for not introducing synchronized.  I concur, and will back-port this change to JTS.

Thanks for pointing out some of the most critical places where JTS code should be rewritten to use CoordinateSequences.  I'll make a JTS issue for this.

Issue #184, opened 2017-08-21
Shapefile IO refactoring


xivk:
I want to change the way the shapefile read/write works now:

- [ ] Have a package for this only without dragging along the rest. Now you also get, Postgis and MsSqlSpatial using the NetTopologySuite.IO package. Perhaps even with only one dependency, NetTopologySuite.IO and nothing else.
- [ ] Refactor to support some netstandard version.
- [ ] Remove code that has nothing to do with shapefile writing/reading.

Anyone have any feedback? Should I do this outside of the main NTS project, create a project/package in a new repo?

airbreather:
> Have a package for this only without dragging along the rest. Now you also get, Postgis and MsSqlSpatial using the NetTopologySuite.IO package. Perhaps even with only one dependency, NetTopologySuite.IO and nothing else.

Can you explain what you mean further, please?  The NetTopologySuite.IO package only depends on NetTopologySuite (and thus GeoAPI).

Or are you suggesting that there should be some kind of additional package NetTopologySuite.IO.ShapeFile or NetTopologySuite.IO.GeoTools or something, which **only** depends on GeoAPI?  If so, I like this idea, if you can rip out its uses of other stuff from NetTopologySuite proper.

From memory, I think the only major thing the NetTopologySuite.IO.GeoTools uses from NetTopologySuite proper is the stuff from the NetTopologySuite.Features namespace, so it sounds possible.  Doing it in a backwards-compatibility would be a nuisance.

> Refactor to support some netstandard version.

I agree that shapefile reading / writing support in .NET Standard is important enough to justify.  Unless you get really aggressive with it, I think the lowest you can do is .NET Standard 2.0 (especially for the use case that I think you have in mind, which requires all the APIs that make me say that).

> Remove code that has nothing to do with shapefile writing/reading.

Looking at NetTopologySuite.IO.GeoTools, I don't really think there's all that much of this... am I looking in the wrong place?

xivk:
There is IO.GeoTools but it has a dependency on IO.ShapeFile. I think we need to define what does what here too because it's a bit confusing having this:

- IO.ShapeFile: ?
- IO.GeoTools: Reads/Writes shapfiles.

Not suggesting only depending on GeoAPI, I don't think that's a good choice because we need the geometries in NTS.

This code I don't think is IO and dumps a shapefile into a database:

[...]

PR #200, opened 2017-10-25
Add .NET Standard 2.0 Target

airbreather:
In summary, this PR adds the .NET Standard 2.0 target framework to all projects that already target .NET Standard 1.x, and also to `NetTopologySuite.IO.GeoTools` and `NetTopologySuite.IO.ShapeFile.Extended`, which exposes the "familiar" shapefile reading / writing functionality of NetTopologySuite to all platforms that implement .NET Standard 2.0 (see [this matrix](https://github.com/dotnet/standard/blob/master/docs/versions.md) for which platforms those are).

It does so without (knowingly) introducing any breaking API changes that would be visible to downstream consumers on the full .NET Framework, and with only one (known) relevant public API difference between the .NET Standard 2.0 and "fat" .NET Framework targets (namely, the change in the single modified C# source code file).

This intersects a lot with #184, as @xivk has an alternative suggestion for how to bring shapefile reading / writing to other target frameworks (most interestingly, .NET Standard 1.3).  I've shared some opinions in the comments of that issue about that idea.

Because of the intersection with #184, I'm going to hold off a bit on merging this until others have an opportunity to share their opinions on this.
1. If this goes out there "soon", and then we bring in the breaking API changes proposed in the most recent concrete solution proposed in #184, then those breaking changes can affect any downstream consumers who would have begun to use the new things added here, which I think just looks bad.
2. If, on the other hand, those breaking changes are added before we do something like this PR, then downstream consumers on frameworks other than "fat" .NET Framework would not be impacted by any of those breaking changes because they would only develop against the aftermath.
3. Finally, if we reject those breaking changes for now, then this would be "perfectly fine" as-is, apart from the fact that I still really hate the "design" of the shapefile reading / writing API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: csharp
 mono: latest
-solution: NetTopologySuite.sln
+dotnet: 2.0.0
 dist: trusty
 install:
 - nuget install NUnit.Runners -Version 2.6.4 -OutputDirectory testrunner
--- a/NTS.nuspec
+++ b/NTS.nuspec
@@ -28,24 +28,24 @@
 		<dependencies>
 			<!-- traditional .NET framework targets -->
 			<group targetFramework=".NETFramework3.5-Client">
-				<dependency id="GeoAPI" version="[1.7.5-pre006, )" />
+				<dependency id="GeoAPI" version="[1.7.5-pre009, )" />
 			</group>
 
 			<group targetFramework=".NETFramework4.0-Client">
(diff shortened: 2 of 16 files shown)


airbreather:
After making this change, [this](https://gist.github.com/airbreather/0937a51d8949e467e1bc61be7c970e3f) works in a `netcoreapp2.0` app.

airbreather:
This would happen to fix #201 at the same time, which for me tilts the balance in favor of just merging it and accepting the risk of a breaking change depending on how #184 goes.

PR #203, opened 2017-11-01
Fix IndexOutOfRangeException in BasicSegmentString

AsgerPetersen:
`BasicSegmentString.IsClosed` always fails with an IndexOutOfRangeException


DGuidi:
thanks for this

PR #214, opened 2018-02-27
Resolves#213

grisha0088:
Now if it is nonhomogenous igeometry, we compare all geometries in collection.


airbreather:
JTS [doesn't currently do this](https://github.com/locationtech/jts/blob/42a92b4f2c0e077324cde375cc5a52557da16021/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L1094-L1100), so you'd probably have better luck getting it to happen there.

On its own merits, I think this seems too naive.  Simply changing the order of geometries within a collection would cause it to fail when it should pass.

Issue #216, opened 2018-03-02
GeoJSON feature ID is lost in parsing when feature has no properties


dalbani:
As far as I understand, given that [FeatureConverter](https://github.com/NetTopologySuite/NetTopologySuite/blob/develop/NetTopologySuite.IO/NetTopologySuite.IO.GeoJSON/Converters/FeatureConverter.cs#L151) relies on the `Attributes` property of a `Feature` to store its ID, the value is ignored during parsing when no `properties` where present in the original GeoJSON.
Example:
```json
{
  "id": "featureID",
  "type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": [42, 42]
  }
  // no "properties" here!!
}
```
In this case, AFAIK, there's no way to retrieve the original ID from the deserialized `Feature` object, because `Attributes` is null and has not been filled in with the ID.

Tested with latest 1.15-pre version, with .NET Core 2.0.

I see that an issue was opened in 2016 about GeoJSON feature IDs: https://github.com/NetTopologySuite/NetTopologySuite/issues/83.
A workaround was given, but it doesn't work in the case I mention here.

I suppose a fix would be to always create an empty `AttributesTable` — but then you lose the ability to determine if no `properties` were present in the original GeoJSON.

Compared to other implementations of GeoJSON parses, the use of the `Attributes` property to store the ID looks like a hack.
Wouldn't it be possible to have a dedicated property in the `Feature` class??

airbreather:
This issue was moved to NetTopologySuite/NetTopologySuite.IO.GeoJSON#6

PR #219, opened 2018-03-07
Always keep track of GeoJSON feature ID when parsing

dalbani:
Fixes the issue described in #216.
If this change is deemed acceptable, I shall add a unit test.


FObermaier:
We are in the process of moving IO projects into separate repositories.
Could you please create your pull-request on [NetTopologySuite.IO.GeoJSON](https://github.com/NetTopologySuite/NetTopologySuite.IO.GeoJSON).

A unit test is mandatory :)

PR #221, opened 2018-03-13
Added non-seekable stream support

YohDeadfall:
We've faced an issue with reading data from a non-seekable stream using `PostGisReader` because of these lines:

https://github.com/NetTopologySuite/NetTopologySuite/blob/543b56b20c8c83df9d7ae12cf2ac3a15590c349f/NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs#L97-L104

The solution is to have a `BinaryWriter` which allows to change byte order on the fly. The `PostGisReader` reads the actual byte order in the `Read` method and sets it to `ConfigurableBinaryReader.EncodingType`.

### DIFF
--- a/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
@@ -8,7 +8,7 @@ namespace NetTopologySuite.IO
     {
         public override IGeometry Read(Stream stream)
         {
-            using (var reader = new BiEndianBinaryReader(stream, byteOrder))
+            using (var reader = new BiEndianBinaryReader(stream))
             {
                 IGeometry geometry = Read(reader);
                 int srid = -1;
--- a/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
@@ -8,7 +8,6 @@ namespace NetTopologySuite.IO
     {
         public override IGeometry Read(Stream stream)
         {
-            var byteOrder = (ByteOrder)stream.ReadByte();
             using (var reader = new BiEndianBinaryReader(stream, byteOrder))
             {
                 IGeometry geometry = Read(reader);
--- a/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
@@ -9,7 +9,7 @@ namespace NetTopologySuite.IO
         public override IGeometry Read(Stream stream)
         {
             var byteOrder = (ByteOrder)stream.ReadByte();
-            using (var reader = new ConfigurableBinaryReader(stream, byteOrder))
+            using (var reader = new BiEndianBinaryReader(stream, byteOrder))
             {
                 IGeometry geometry = Read(reader);
                 int srid = -1;
--- a/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
@@ -9,7 +9,7 @@ namespace NetTopologySuite.IO
         public override IGeometry Read(Stream stream)
         {
             var byteOrder = (ByteOrder)stream.ReadByte();
-            using (BinaryReader reader = byteOrder == ByteOrder.BigEndian ? new BEBinaryReader(stream) : new BinaryReader(stream))
+            using (var reader = new ConfigurableBinaryReader(stream, byteOrder))
             {
                 IGeometry geometry = Read(reader);
                 int srid = -1;
--- a/NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs
+++ b/NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs
@@ -94,7 +94,7 @@ namespace NetTopologySuite.IO
         /// <returns></returns>
         public IGeometry Read(Stream stream)
         {
-            using (var reader = new ConfigurableBinaryReader(stream))
+            using (var reader = new BinaryReader(stream))
                 return Read(reader);
         }
 
@@ -108,9 +108,9 @@ namespace NetTopologySuite.IO
         /// </summary>
         /// <param name="reader"></param>
         /// <returns></returns>
-        protected IGeometry Read(ConfigurableBinaryReader reader)
+        protected IGeometry Read(BinaryReader reader)
         {
-            reader.EncodingType = (ByteOrder)reader.ReadByte();
+            ((BiEndianBinaryReader)reader).Endianess = (ByteOrder)reader.ReadByte();
 
             var typeword = reader.ReadInt32();
 
@@ -293,7 +293,7 @@ namespace NetTopologySuite.IO
         /// </summary>
         /// <param name="reader">The binary reader.</param>
         /// <param name="container">The container for the geometries</param>
-        protected void ReadGeometryArray<TGeometry>(ConfigurableBinaryReader reader, TGeometry[] container)
+        protected void ReadGeometryArray<TGeometry>(BinaryReader reader, TGeometry[] container)
             where TGeometry : IGeometry
         {
             for (var i = 0; i < container.Length; i++)
@@ -306,7 +306,7 @@ namespace NetTopologySuite.IO
         /// <param name="reader">The binary reader.</param>
         /// <param name="factory">The geometry factory to use for geometry creation.</param>
         /// <returns>The MultiPoint</returns>
-        protected IMultiPoint ReadMultiPoint(ConfigurableBinaryReader reader, IGeometryFactory factory)
+        protected IMultiPoint ReadMultiPoint(BinaryReader reader, IGeometryFactory factory)
         {
             int numGeometries = reader.ReadInt32();
             var points = new IPoint[numGeometries];
@@ -320,7 +320,7 @@ namespace NetTopologySuite.IO
         /// <param name="reader">The binary reader.</param>
         /// <param name="factory">The geometry factory to use for geometry creation.</param>
         /// <returns>The MultiLineString</returns>
-        protected IMultiLineString ReadMultiLineString(ConfigurableBinaryReader reader, IGeometryFactory factory)
+        protected IMultiLineString ReadMultiLineString(BinaryReader reader, IGeometryFactory factory)
         {
             int numGeometries = reader.ReadInt32();
             var strings = new ILineString[numGeometries];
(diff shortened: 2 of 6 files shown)


NetTopologySuite/IO/ConfigurableBinaryReader.cs
```diff
@@ -0,0 +1,220 @@
...
+    /// </summary>
+    /// <remarks>
+    /// While <see cref="ConfigurableBinaryReader" /> extends <see cref="BinaryReader" /> 
+    /// adding methods for reading integer values (<see cref="ConfigurableBinaryReader.ReadInt32" />)
+    /// and double values (<see cref="ConfigurableBinaryReader.ReadDouble" />) in the specified format, 
+    /// this implementation overrides methods, such <see cref="BinaryReader.ReadInt32" /> 
+    /// and <see cref="BinaryReader.ReadDouble" /> and more, 
+    /// for reading values in the specified format.
+    /// </remarks>
+    public class ConfigurableBinaryReader : BinaryReader
```
airbreather: Can we name this something that more specifically identifies byte order as the thing you're "configuring"?

NetTopologySuite/IO/ConfigurableBinaryReader.cs
```diff
@@ -0,0 +1,220 @@
...
+
+        /// <summary>
+        /// Reads a string from the current stream. 
+        /// The string is prefixed with the length, encoded as an integer seven bits at a time.
+        /// </summary>
+        /// <returns>The string being read.</returns>
+        /// <exception cref="T:System.ObjectDisposedException">The stream is closed. </exception>
+        /// <exception cref="T:System.IO.IOException">An I/O error occurs. </exception>
+        /// <exception cref="T:System.IO.EndOfStreamException">The end of the stream is reached. </exception>
+        [Obsolete("Not implemented")]
```
airbreather: Can we also add `, error: true` to these two `[Obsolete]` attributes so it outright fails to compile if someone tries to use these?

airbreather: Obviously won't help with people who call these virtual methods virtually, but it's the best we can do...

NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs
```diff
@@ -114,11 +108,9 @@ public IGeometry Read(Stream stream)
         /// </summary>
         /// <param name="reader"></param>
         /// <returns></returns>
-        protected IGeometry Read(BinaryReader reader)
+        protected IGeometry Read(ConfigurableBinaryReader reader)
         {
-            // Dummy read, just for bytes compatibility.
-            // The byte order is determined only once.
-            reader.ReadByte();
+            reader.EncodingType = (ByteOrder)reader.ReadByte();
```
airbreather: By itself this seems fine to me, but it also seems odd that the regular ol' `WKBReader` (and, by extension, `MsSqlSpatialReader`) just reads the byte order once at the very start and then ignores the byte order of nested geometries.  Perhaps we should have those also use this?

airbreather: After doing so, we could also consider marking `BEBinaryReader` obsolete, since those would be the last two things that use it.

NetTopologySuite/IO/ConfigurableBinaryReader.cs
YohDeadfall: Changed the comment. Is it okay now? Or did you mean the class name?

airbreather: Yes the class name is what I was referring to.

NetTopologySuite/IO/WKBReader.cs
```diff
@@ -245,12 +224,12 @@ protected IGeometry Read(BinaryReader reader)
         /// 
         /// </summary>
         /// <param name="reader"></param>
-        private void ReadByteOrder(BinaryReader reader)
+        private void ReadByteOrder(ConfigurableBinaryReader reader)
         {
             reader.ReadByte();
```
airbreather: Ahh, when I'd made my original comment, I meant that it seems like at least this line should change to set `reader.EncodingType` here, since the PostGis reader's version would do the same.

NetTopologySuite/IO/ConfigurableBinaryReader.cs
YohDeadfall: Since it's a binary reader the only thing which can be changed is byte order. I don't take the `encoding` into account because we don't need it in that case. But yep, probably the name should be changed... And currently I have no idea what would be better. @roji?

roji: Hmmm... The ability to handle both endian methods could be expressed by BiEndianBinaryReader. But the main point of this class seems to add support for reading int/double, so maybe something like NumericBinaryReader?

FObermaier: `EndianessAware[Binary]Reader` dreadfully long but describes it.
I'd say this class can be `internal`

NetTopologySuite/IO/ConfigurableBinaryReader.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// <param name="input">The supplied stream.</param>
+        /// <param name="encoding">The character encoding.</param>
+        /// <exception cref="T:System.ArgumentNullException">encoding is null. </exception>
+        /// <exception cref="T:System.ArgumentException">The stream does not support reading, the stream is null, or the stream is already closed. </exception>
+        public ConfigurableBinaryReader(Stream input, Encoding encoding) : base(input, encoding) { }
+
+        /// <summary>
+        /// Encoding type
+        /// </summary>
+        public ByteOrder EncodingType { get; set; }
```
FObermaier: public ByteOrder `Endianess` { get; set; }

NetTopologySuite/IO/ConfigurableBinaryReader.cs
FObermaier: Why `[Obsolete]` in the first place, just call the base function if Endianess is little, otherwise throw NotSupportedException();
This is a tweak for our problem, not for the world.

NetTopologySuite/IO/ConfigurableBinaryReader.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// Reads a decimal value from the current stream 
+        /// and advances the current position of the stream by sixteen bytes.
+        /// </summary>
+        /// <returns>
+        /// A decimal value read from the current stream.
+        /// </returns>
+        /// <exception cref="T:System.ObjectDisposedException">The stream is closed. </exception>
+        /// <exception cref="T:System.IO.IOException">An I/O error occurs. </exception>
+        /// <exception cref="T:System.IO.EndOfStreamException">The end of the stream is reached. </exception>
+        [Obsolete("Not implemented")]
```
FObermaier: dito

NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs
FObermaier: I'd say leave the signature of the functions as they were before (BinaryReader).
Add a type check (Contract? or check for argument being the required type) and at the point where we read the encoding/endianess cast to the specific reader.

YohDeadfall: Same for `WKBReader`?

NetTopologySuite/IO/ConfigurableBinaryReader.cs
YohDeadfall: The name came from `WKBWriter`. It has the `EncodingType` field. Change it anyway?

NetTopologySuite/IO/ConfigurableBinaryReader.cs
YohDeadfall: I would like to prefer `BiEndianBinaryReader` because it's shorter and has the same meaning as `EndianessAwareBinaryReader`. Fine?

It can't be internal since `PostGisReader` uses it.

NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs
FObermaier: I did not notice you changed them, too, if so, please do that as for PostGisReader.
Or did I miss your point?

NetTopologySuite/IO/ConfigurableBinaryReader.cs
FObermaier: I don't necessarily connect `EncodingType` to byte order or endianess.
I'd say either `Endianess` or `ByteOrder` is the right word.

NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
```diff
@@ -9,7 +9,7 @@ public class MsSqlSpatialReader : WKBReader
         public override IGeometry Read(Stream stream)
         {
             var byteOrder = (ByteOrder)stream.ReadByte();
```
airbreather: Isn't this wrong now that the `Read(reader)` we do later will also try to read the byte order?

NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs
```diff
@@ -116,9 +110,7 @@ public IGeometry Read(Stream stream)
         /// <returns></returns>
         protected IGeometry Read(BinaryReader reader)
         {
-            // Dummy read, just for bytes compatibility.
-            // The byte order is determined only once.
-            reader.ReadByte();
+            ((BiEndianBinaryReader)reader).Endianess = (ByteOrder)reader.ReadByte();
```
airbreather: Won't this cast fail because we don't create it as `BiEndianBinaryReader` on line 97?

airbreather: Without changing line 97 to actually create it as `BiEndianBinaryReader`, that wouldn't actually help because it would just never set the property.

NetTopologySuite.IO/NetTopologySuite.IO.MsSqlSpatial/MsSqlSpatialReader.cs
YohDeadfall: Thanks! Fixed.

NetTopologySuite.IO/NetTopologySuite.IO.PostGis/PostGisReader.cs
YohDeadfall: My mistake. Fixed now. Do you have tests for geometry readers?

FObermaier:
There is a [test project for all NTS.IO](https://github.com/NetTopologySuite/NetTopologySuite/tree/develop/NetTopologySuite.IO/NetTopologySuite.IO.Tests) projects. You will have to  adjust the connection string.

YohDeadfall:
@airbreather When will you able to merge the PR and publish a new version of the NuGet package?

airbreather:
@YohDeadfall that's a question for @FObermaier; I don't feel comfortable merging this myself since it touches stuff I've never used, myself.

YohDeadfall:
Not a problem. I just thought that he didn't have permissions to do that (:

FObermaier:
@YohDeadfall , sorry but we are in the process of seperating the NTS project into different repositories, one of them being `https://github.com/NetTopologySuite/NetTopologySuite.IO.PostGis`. As @xivk has already put quite some effort into this I'd rather merge this into that repository.
Additionally @roji  has suggested to put pre-release packages on myget. I don't know how far that has gotten.

roji:
> @YohDeadfall , sorry but we are in the process of seperating the NTS project into different repositories, one of them being https://github.com/NetTopologySuite/NetTopologySuite.IO.PostGis. As @xivk has already put quite some effort into this I'd rather merge this into that repository.

Good to hear you're going down that road, it makes a lot of sense...

Thanks for all the work on NTS, I'm really happy to see that there's a full-featured spatial library for .NET that we can consider standard.

roji:
Just to send a reminder on this.

FYI the next version of Npgsql, the PostgreSQL .NET driver, will be shifting to NetTopologySuite as it's recommended/first-class spatial access method, so that NetTopologySuite types will be readable and writable directly from and to PostgreSQL via the ADO.NET API.

I appreciate the great work you guys are doing, I hope you're able to merge this and release 1.15.0 in the near future.

PR #228, opened 2018-04-22
Some code style fixes

roji:
While going over the NTS source code, I encountered several coding style issues - modern C# 7 features aren't used (e.g. expression-bodied properties and methods), indentation and newlines are frequently inconsistent...

This is a first PR that addresses some of these issues. If you're interested I can help out a bit more on cleanup.

(that's what happens when you're an obssessive-compulsive personality...)

PR #230, opened 2018-04-27
Normalized line endings to LF

roji:
While working on some styling issues in #228, I noticed that line endings were inconsistent in the repo, and that there was no git enforcement configured. In other words, some users can currently contribute code with CRLF and others with LF, leading to more inconsistency. This creates false changes as code goes back and forth between users etc.

The git standard is to have all files in the repo with LF line endings, and have git normalize all files upon commit (see [the gitattributes docs](https://git-scm.com/docs/gitattributes) and [the github docs](https://help.github.com/articles/dealing-with-line-endings/)).

This is a somewhat painful step since all files are modified, but once done line ending confusion is gone forever.

* Activated git normalization of line endings to LF via a new .gitattributes file.
* Normalized all existing files (git add --renormalize .)
* Added editorconfig to set indentation at 4 spaces and some other whitespace-related settings.


airbreather:
Since this is a largely whitespace-only change, when reviewing this on GitHub, I recommend adding `?w=1` to the review URL, as follows:

[`https://github.com/NetTopologySuite/NetTopologySuite/pull/230/files?w=1`](https://github.com/NetTopologySuite/NetTopologySuite/pull/230/files?w=1)

That will tell GitHub to ignore whitespace-only changes when presenting files for review.

PR #232, opened 2018-04-29
Various code style fixes

roji:
A pack of C# coding changes to make the project more modern and easier to work with. The only thing that IMHO may be a bit controversial is the shift to using `var` everywhere; the project as is mixes both var and non-var, unless you guys have special feelings about this it seems to be the standard/right thing to do. If you object let me know and I'll remove that commit.

These changes were done automatically with Rider/Resharper but it's a good idea to quickly scan just to make sure everything is OK.


airbreather:
Can any / all of these style preferences (`var` vs. explicit type, when to favor expression-bodied members, when to favor using framework types vs. their C# builtin aliases) be expressed through `.editorconfig`?  Whatever we settle on for style preferences, if VS2017 / Rider / Visual Studio Code / etc. aren't going to help "enforce" it (squiggly lines when we do the wrong thing), then it seems like we're fighting a losing battle.

And if they're in `.editorconfig`, then in theory our IDEs should be able to give us a button that lets us conform the whole solution in one go, automatically.

Pretty sure "don't leave whitespace at the end of lines" and "where possible, use auto-properties instead of explicit fields that the property simply wraps" are uncontroversial enough that they shouldn't really need settings, though.

NetTopologySuite/Geometries/Geometry.cs
```diff
@@ -2256,14 +2196,7 @@ private static IPoint CreatePointFromInternalCoord(Coordinate coord, IGeometry e
         /// 
         /// </summary>
         /// <returns></returns>
-        public virtual bool IsRectangle
-        {
-            get
-            {
-                // Polygon overrides to check for actual rectangle
```
airbreather: There was a comment here that I kinda liked.

NetTopologySuite/Geometries/MultiLineString.cs
```diff
@@ -132,18 +115,7 @@ public bool IsClosed
         //    }
         //}
 
-       public override IGeometry Boundary
-        {
-            get
-            {
-                return (new BoundaryOp(this)).GetBoundary();
-                //if(IsEmpty)
```
airbreather: Not sure what this commented-out code was for, but it's gone now.

NetTopologySuite/Utilities/RToolsUtil/StreamTokenizer.cs
```diff
@@ -193,28 +205,8 @@ public bool ParseNumbers
 		/// </summary>
 		public bool ParseHexNumbers
 		{
-			get 
-			{ 
-				return(parseHexNumbers);
-//				for (int i = 'A'; i <= 'F'; i++)
```
airbreather: More commented-out code that went away

roji:
Thanks for giving this time @airbreather. Yeah, I actually have some problems building the project here (especially because of the old target frameworks such as `net35-client`), so I was actually waiting for Travis to finish the build - but you beat me to it. I can work on making the PR pass the travis build before you spend any more time - as you wish. Looking at the build, I can see one test failure that seems to be a result of some bad `var` changes, I'll take a look at it.

Regarding your question .editorconfig question, yes - there are some newish standard C# options for .editorconfig which are supposed to be respected by VS, R# and other .NET IDEs, take a look at [this docs page](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference). 

In the existing .editorconfig that has already been added, there's already an IDE instruction to trim trailing whitespace, but no C#-specific things. If we decide to, we can indeed set up .editorconfig to flag an error if, say `var` isn't used; and I think similar options exist for some of the other stuff. This is a tad extreme - note that VS, Resharper and Rider already provide "suggestions" for all of the stuff I proposed above (convert to auto-property, expression-bodied properties) but they aren't considered errors. Up to you guys to decide how you want to do it.

Yeah, I think that after these changes we can do one more pass, redoing all indentation to 4 spaces (I noticed the inconsistencies as I was working).

Let me know how you want to proceed. If you want me to submit separate PRs for each commit, we can do it that way (although it would be a bit long).

Finally, note that all this work probably needs to be repeated for each and every additional repo that was split off from the main one... Once we complete the work on the main repo I can try to submit similar PRs for the other repos.

roji:
> Similar to the previous review, this is just looking at 30bfb30.
> 
> I personally really don't like how this commit seems to have removed blank lines that were used to separate elements. Everything looks squished together now.

Oops, this was definitely unintended - thanks for catching it. If you'd like I can amend it.

Note that you can also cherry-pick specific commits which are OK and merge them as-is, and leave others for me (or you) to work on.

FObermaier:
Browsing through @roji commits I'd like to note the following:

* C# expression-bodied properties 8ccdd6c    
no objections, as this is just syntax, isn't it. As @airbreather pointed out comments and commented debug purpose snipplets shouldn't be removed though.
* Strip all whitespace at end of line 30bfb30   
no objections either, there even might be some whitespace inhomogenity, has anyone checked this?
* C# refactor to auto-properties 3f571d2   
this is problematic for the maintenance of the code base and keeping it synchronized with JTS, as class members are being removed. I'd prefer to leave this out.
* Use var everywhere
I'm a big fan of this but I know that @DGuidi  is not.
* Use C# simple type names
No objections

Generally I object to anything that somehow rearranges the code (Sort by construction, property or method).

roji:
@FObermaier thanks - everything you say seems reasonable. I'll remove the auto-properties refactor and fix everything that needs fixing. I'll also look into homogenizing indentation to 4 spaces - I know there's quite a bit of mixed indentation in the code base.

However, you guys will need to make a decision on var :)

airbreather:
> However, you guys will need to make a decision on var :)

Regarding `var`, my personal ideal is really close to what can be done with a few [`.editorconfig` settings](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference#var):

1. `csharp_style_var_for_built_in_types = false`
    - `var i = 5;` just looks silly to me
2. `csharp_style_var_when_type_is_apparent = true`
    - some type names are really long, and it doesn't really seem to harm anything to write `var dict = new Dictionary<string, SomeOtherReallyLongTypeName>();`
3. `csharp_style_var_elsewhere = true`
    - here, I'm perfectly fine either way, and there are cases on both sides where I would specifically favor "the other one" no matter which we pick, but I do see the value in picking one and sticking to it.

To me, though, what's more important than the exact rules is that we configure tools so that I can push a button to make everything conform instead of having to spend serious manual time.

FObermaier:
>To me, though, what's more important than the exact rules is that we configure tools so that I can push a button to make everything conform instead of having to spend serious manual time.

Definetly yes 👍. If all this can be handled using/configuring `.editorconfig` than we should do that.   

I have no objections to the rules you suggest for the usage of `var`. I could live with `var i = 5;`, too, though. @DGuidi is very busy atm, maybe he can give is consent anyway :-).?

airbreather:
For these commits, I've merged either the commits themselves or rough equivalents to them:
- 8ccdd6c (with comments restored in a separate commit)
- 30bfb30 (redid it less aggressively, since important blank lines had been removed)
- 1712a2a (redid it, plus updated `.editorconfig` so it identifies violations in the future)

I have done nothing for the following two, since there's not a consensus:
- 3f571d2 (auto-properties aren't there in Java, so this would be a maintainability headache... we'll probably never do this)
- 26dba84 (to `var` or not to `var`: that is the question... we'll need to pick one, and we haven't yet)

roji:
@airbreather I've been wanted to amend this PR, thanks for doing it :)

I think that similar changes need to be done for the other repos which have been split off, but I don't have much time at the moment to work on it...

FObermaier:
@airbreather , @roji : like I mentioned previously I have no objections to the rules @airbreather suggested regarding the use of `var`:
```
csharp_style_var_for_built_in_types = false
csharp_style_var_when_type_is_apparent = true
csharp_style_var_elsewhere = true
```
Maybe `csharp_style_var_when_type_is_apparent = true` is redundant?

Since @DGuidi  has not complained I think we can go ahead.

airbreather:
I've merged in a fix for the tabs-vs-spaces issue.  The only tab characters that remain in `*.cs` files look like a direct copy-paste output from an "Error List" window, so I left them:
https://github.com/NetTopologySuite/NetTopologySuite/blob/2204937b03ae283146b53f5a32f920ef74aacb29/NetTopologySuite.Samples.Console/Tests/Various/Issue174TestFixture.cs#L43-L50

I've given the `var` thing a start in branch [`style-fixes/consistent-var`](https://github.com/NetTopologySuite/NetTopologySuite/tree/style-fixes/consistent-var).  One thing that bothers me a little, browsing through the changed files, is that there are some things that would be really nice for the auto-fixer to fix for us, but which it can't (as a rule).  For example:
https://github.com/NetTopologySuite/NetTopologySuite/blob/3741869f4612c5c7d2564cbd1c1f11b856333231/NetTopologySuite.Lab/Geometries/Utilities/GeometryEditorEx.cs#L244

Technically, "fixing" that would actually slightly change the semantics of that line.  The bothersome part is that the "fixed" semantics would actually be preferable in this case, because invoking the methods through the interface like this will tend to make the JIT generate slower machine code (never faster, as far as I'm aware: always just "slower" or "the same").

I'll do a **bit** of hunting to make these better in the main library where I know it's safe, but otherwise I'll leave it as-is.

airbreather:
To the best of my knowledge and abilities, I've updated `develop` so that it now follows all of the agreed-upon rules, including `.editorconfig` updates to flag errors on code that violates these rules ([they can't fail the build right now](https://github.com/dotnet/roslyn/projects/18), but it seems to be the best we can do).

I've also banished all tab characters from `*.cs` files with the exception of the one block that I linked in an above comment.

I believe that this covers everything that this PR set out to do, so I'm closing this PR now.  Thanks @roji for getting the ball rolling on this!

roji:
Great to see this cleaned up, a clean code base is an important step in any project. And sorry I don't have more time to help out at the moment... I hope that will change in the future and I can perhaps contribute a bit more.

PR #240, opened 2018-05-06
#239: all tests run with .NET Core now

### LINKED ISSUE #239: Change tests to use .NET Core (requires NUnit update)
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of NetTopologySuite
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/NetTopologySuite/NetTopologySuite/issues) and [closed](https://github.com/NetTopologySuite/NetTopologySuite/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
The tests all use .NET Fat 4.5.

Now that we've moved all the Windows-specific junk out of the main NetTopologySuite solution, there's no compelling reason to keep using Mono, instead of the normal .NET Core SDK, to run the tests on Linux-based systems (this is annoying not just for Travis-CI, but also any other contributor who uses Linux as their primary development platform).

Furthermore, several other issues, most recently #227, have made it so that there should be no reason why any tests may be expected to fail as-is when run against the .NET Standard 2.0 target framework.

Let's switch the NTS test project(s) to use .NET Core 2.0.  Doing so properly (i.e., in a way that's compatible with the VS2017 test runner) seems to require us to update NUnit to 3.x+.

airbreather:
### Prerequisites

### Description
Resolves #239 (see that issue for the rationale).

Switching all the test projects over to target `netcoreapp2.0`.  This required getting rid of some old `using` blocks and the `System.Drawing` stuff from `VoronoiExample.TestVoronoi`.

I was kind-of hoping to be able to drop the Mono stuff from `.travis.yml` at the same time, but it's looking more and more like that won't be possible until we can drop `net35-client`, `net40-client`, and `net403-client` (if we did drop those three, then we could probably add a `PackageReference` to [`Microsoft.TargetingPack.NETFramework.v4.5`](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.TargetingPack.NETFramework.v4.5) and allow building on Linux with minimal dependencies).

NUnit has made some breaking changes since 2.x:
- `[TestFixtureSetUp]` has been renamed to `[OneTimeSetUp]`
- `[TestFixtureTearDown]` has been renamed to `[OneTimeTearDown]`
- `[Ignore]` now **requires** a `reason` (so I just gave a dummy one).
- `[ExpectedException]` is gone (so I just wrapped the old method in an `Assert.Throws<...>`).

`ReleaseLinux` solution build configuration is redundant with `Release` now, so it's gone.


airbreather:
@FObermaier @DGuidi I think this looks OK, thoughts?

I've run the console apps a little bit (0 failures for TestRunner.Console in 'default', 20 failures in 'other', which matches what I get untouched).  There are several samples in `NetTopologySuite.Samples.Console` that I haven't run, but it does all build and the NUnit tests all pass (ignoring the one marked `FailureCase`), so I figured that's good enough.

FObermaier:
Thanks @airbreather , for all the effort you put into this library

PR #255, opened 2018-08-27
Fix PackedCoordinateSequenceFactory.Ordinates

bricelam:


airbreather:
I wonder if we should actually use a backing field for it, so that we can distinguish between `XYZ` and `XYM` (obviously means adding more c'tor overloads to match the corresponding factory method)...

bricelam:
`PackedCoordinateSequence` seems to handle it all a lot better... This at least unblocks using XYZM in WKBReader (see [this line](https://github.com/NetTopologySuite/NetTopologySuite/blob/2579161109a4c123464163c7a6b688512a34b64f/NetTopologySuite/IO/WKBReader.cs#L591)) and possibly other readers...

airbreather:
Thank you for your contribution; I'll look into my other comments separately from this.

PR #290, opened 2019-03-02
Port JTS' commits up to JTS 1.16.0 and target .NET Standard 2.0 only

### LINKED ISSUE #265: Update to JTS 1.16.0
[JTS 1.16.0](https://github.com/locationtech/jts/releases/tag/jts-1.16.0)

There's at least one intentional breaking change in there (locationtech/jts@c709631 removed the `(int, int)` constructor that [we still have](https://github.com/NetTopologySuite/NetTopologySuite/blob/c94898f6a6282f6d8bb4fa65062ac59324108b85/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequenceFactory.cs#L52-L61)), so we would technically violate SemVer if we released it as NTS 1.16, but we have a history of way bigger SemVer violations, so it shouldn't be the end of the world.

Here's the list of commits to consider.

- [x] locationtech/jts@4d5b984 (JTS-specific, meta)
- [x] locationtech/jts@7da8065 (JTS-specific, meta)
- [x] locationtech/jts@2387a45 (entirely in GeoAPI)
- [x] locationtech/jts@d13f230 (f8f20d1)
- [x] locationtech/jts@c709631 (bba1334)
- [x] locationtech/jts@868c34b (00d97f4)
- [x] locationtech/jts@8c5db26 (ce13e8d1)
- [x] locationtech/jts@8ef27d4 (5161d65)
- [x] locationtech/jts@e302001 (18aad66)
- [x] locationtech/jts@06e0424 (ed5be57)
- [x] locationtech/jts@17e9b05 (d00fed1)
- [x] locationtech/jts@bd4ca6e (ca328e4)
- [x] locationtech/jts@412bf17 (JTS-specific, meta)
- [x] locationtech/jts@7f8cde4 (JTS-specific, meta)
- [x] locationtech/jts@b1422b4 (2878d59)
- [x] locationtech/jts@2477de9 (1c9cded)
- [x] locationtech/jts@0b75fe7 (d1a5ed1)
- [x] locationtech/jts@19e6f32 (82acbe3)
- [x] locationtech/jts@b464d54 (f4afb40)
- [x] locationtech/jts@58ba543 (52d1f7e)
- [x] locationtech/jts@09e3554 (9715e0d)
- [x] locationtech/jts@ef26db0 (35cf416)
- [x] locationtech/jts@88e967b (0d5386c)
- [x] locationtech/jts@a1009c1 (e219d3c)
- [x] locationtech/jts@3221a31 (nothing to do, we've always used the same GeoAPI enum)
- [x] locationtech/jts@8d433ca (JTS-specific, meta)
- [x] locationtech/jts@f72773a (655db33)
- [x] locationtech/jts@7d7d796 (602f10d)
- [x] locationtech/jts@83b684f (0331f76)
- [x] locationtech/jts@34cf94d (JTS-specific, meta)
- [x] locationtech/jts@d3d91ab (JTS-specific, meta)

airbreather:
Resolves #265 

- Updating to target only `netstandard2.0` lets us skip having Travis-CI install Mono, which means that our CI build times went from >10 minutes to roughly 3 minutes.
- `WKTReader` was using `StreamTokenizer` significantly differently from how Java was using it.
    - Most critically, we were using `StreamTokenizer`'s number parsing whereas JTS [specifically decided not to do that](https://github.com/locationtech/jts/blob/bac855f9fe3a75af523b43f95f4b9a09b4ab5359/modules/core/src/main/java/org/locationtech/jts/io/WKTReader.java#L192-L202), in favor of treating them as words (probably to have `NaN` use the same code path).
    - This was causing some headaches in trying to merge the new changes, so I brought this back to how JTS does it.
- Each individual commit was ported **with** as many bugs as I could identify, so that later commits that fixed those bugs would be recognizable.
- 3d7375aa is to ensure that the original stack trace is preserved when we rethrow the original exception.
    - Otherwise you'd get the original message but lose the line that actually **threw** the exception.
    - As you might imagine, this makes it incredibly hard to debug when the exception is caused by trying to set Z on a coordinate that doesn't support it.
- 59e1d74d fixes strictly JTS issues that I reported here: https://github.com/locationtech/jts/issues/370#issuecomment-468926825
- I'm a bit unsure about why fd999bfa was needed, since it seems weird that ProjNET4GeoAPI would fail... I think we can address that later, though.


airbreather:
Please do not merge this; I would like to make a custom merge commit for this to make sure that the history identifies this series properly and links it to all the other work items that are waiting on this.

FObermaier:
I missed: there must have been a commit to update the version number to v1.16.

airbreather:
> update the version number to v1.16.

Good point, but should we instead update it to 2.0 because of the breaking changes?

We can include a separate "JTS compatibility level" attribute if we want.

DGuidi:
my vote is to maintain **v1.16**, looks clearer to me in JTS/NTS world

airbreather:
Let's discuss version numbering on #298

PR #339, opened 2019-09-06
  Implementation of ConcaveHull using simple Chi-Hull algorithm.

grootstebozewolf:
### Prerequisites

- [ ] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository.

### Description
Computes a 𝛘-shape of a Geometry which is a concave Geometry when the threshold is small enough and a convex geometry with a large threshold that contains all the points in the input Geometry. Uses the algorithm defined in [Efficient generation of simple polygons for characterizing the shape of a set of points in the plane](https://www.sciencedirect.com/science/article/abs/pii/S0031320308001180).


FObermaier:
@grootstebozewolf , thanks for your contribution.
The `ConvexHull` stub in JTS/NTS has bothered me for quite some time, too. That's why I ported 
mourners `concaveman` algorithm (https://github.com/locationtech/jts/pull/427). A decision to accept is has not made so far.

From the discussion, I gather that we should not name it `ConcaveHull` but give it some other name describing the algorithm used. If that is met, I think we can merge.

grootstebozewolf:
Maybe we need a strategy pattern? This algorithm is also known as a 𝛘-shape, or ChiShape.

airbreather:
I just pushed a few changes that bring back the old versions of the existing files, since those don't need to be marked as modified for this anymore after the rename.

I'll do a review of this soon.  Sorry it's taken so long, most of my NTS time has been going into trying to tidy up v2 stuff post-release and respond to user feedback about it.

> Maybe we need a strategy pattern? This algorithm is also known as a 𝛘-shape, or ChiShape.

I think that's where the discussion started on the other issue that @FObermaier linked to earlier in this thread.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+                                              where n != null && n.IsLive()
+                                              select n)
+                    {
+                        Debug.WriteLine(neighbour);
+                        CheckMaxEdgeAndListBorderTriangles(quadEdgeTriangleList, neighbour);
+                    }
+                }
+            }
+            var lineMerger = new LineMerger();
+            var fact = new GeometryFactory();
```
airbreather: This should use `_geom.Factory` instead of making a new one.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+namespace NetTopologySuite.Hull
+{
+    public class ChiShape
+    {
+        private readonly Geometry _geom;
+        private readonly double _tolerance;
+
+        public ChiShape(Geometry geom, double tolerance)
+        {
+            _geom = geom;
```
airbreather:
```csharp
_geom = geom ?? throw new ArgumentNullException(nameof(geom));
```

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+                CheckMaxEdgeAndListBorderTriangles(quadEdgeTriangleList, triangle);
+            }
+
+            while (quadEdgeTriangleList.Count != 0)
+            {
+                var triangle = quadEdgeTriangleList.Last();
+                var edge = triangle.Key;
+                Debug.WriteLine(triangle.Value);
+                quadEdgeTriangleList.Remove(edge);
+                if (edge.Length.CompareTo(_tolerance) >= 0 && RemovalGivesRegularPolygon(triangle.Value))
```
airbreather: :question: shouldn't we be breaking out of the loop once we find an edge such that `edge.Length < _tolerance`?  Once we see the first edge that fails the first condition, then I think all other edges will fail too...

Something like:
```csharp
if (edge.Length < _tolerance)
{
    break;
}

if (!RemovalGivesRegularPolygon(triangle.Value))
{
    continue;
}

var neighbours = triangle.Value.GetNeighbours();
// etc...
```

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+                }
+            }
+            var lineMerger = new LineMerger();
+            var fact = new GeometryFactory();
+            foreach (var triangle in tris)
+            {
+                if (!triangle.IsLive()) continue;
+                for (int i = 0; i < 3; i++)
+                {
+                    if (triangle.GetAdjacentTriangleAcrossEdge(i) == null || !triangle.GetAdjacentTriangleAcrossEdge(i).IsLive())
```
airbreather: nit: repeated call to `GetAdjacentTriangleAcrossEdge` can be eliminated:
```csharp
if (triangle.GetAdjacentTriangleAcrossEdge(i)?.IsLive() != true)
```

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+            var subDiv = builder.GetSubdivision();
+            return subDiv;
+        }
+    }
+
+    internal class QuadEdgeComparer : Comparer<QuadEdge>
+    {
+        public override int Compare(QuadEdge x, QuadEdge y)
+        {
+            return x.Length.CompareTo(y.Length);
```
airbreather: The `IComparer<TKey>` used for a `SortedList<TKey, TValue>` is supposed to give a total ordering.  This only checks the `Length`, which means that if two different `QuadEdge`s happen to have the same length, then only one of them will be allowed to live in the list at a given time.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+{
+    public class ChiShape
+    {
+        private readonly Geometry _geom;
+        private readonly double _tolerance;
+
+        public ChiShape(Geometry geom, double tolerance)
+        {
+            _geom = geom;
+            _tolerance = tolerance;
```
airbreather: Probably also validate that `tolerance` is a finite, positive number.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+        private static IList<QuadEdgeTriangle> ExtractTriangles(QuadEdgeSubdivision subdiv)
+        {
+            var qeTris = QuadEdgeTriangle.CreateOn(subdiv);
+            return qeTris;
+        }
+
+
+        private Geometry ComputeHull(IList<QuadEdgeTriangle> tris)
+        {
+            var quadEdgeTriangleList = new SortedList<QuadEdge, QuadEdgeTriangle>(new QuadEdgeComparer());
```
airbreather: After going through everything else, I'm not sure `SortedList<TKey, TValue>` is the right thing to use here.  I'm favoring an `AlternativePriorityQueue<TPriority, TData>` with a `Dictionary<TKey, TValue>` on the side.

- The way we would have to contort our usage of `SortedList<TKey, TValue>` to eliminate an `O(n²)` bit is unnatural.  We only extract values on one end of the sorted list, which is what priority queues optimize for.
- The criteria we should be using for `.ContainsKey` is different from the criteria we use for sorting, which is incompatible with `SortedList<TKey, TValue>`.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+            foreach (var triangle in tris)
+            {
+                CheckMaxEdgeAndListBorderTriangles(quadEdgeTriangleList, triangle);
+            }
+
+            while (quadEdgeTriangleList.Count != 0)
+            {
+                var triangle = quadEdgeTriangleList.Last();
```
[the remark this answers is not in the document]
airbreather: The best I can come up with, looking at just this part of the code, is this `O(log(n))` bit, which I think looks awful:
```csharp
var edge = quadEdgeTriangleList.Keys.Last();
var triangle = quadEdgeTriangleList[edge];
```

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+                if (!triangle.IsLive()) continue;
+                for (int i = 0; i < 3; i++)
+                {
+                    if (triangle.GetAdjacentTriangleAcrossEdge(i) == null || !triangle.GetAdjacentTriangleAcrossEdge(i).IsLive())
+                    {
+                        lineMerger.Add(triangle.GetEdge(i).ToLineSegment().ToGeometry(fact));
+                    }
+                }
+            }
+            var mergedLine = (LineString)lineMerger.GetMergedLineStrings().FirstOrDefault();
```
airbreather: nit: these two lines look like they're arguing about whether or not the list is actually allowed to be empty
- if the list **can** be empty, then `if (mergedLine.IsRing)` will throw `NullReferenceException`
- if the list **can't** be empty, then `.FirstOrDefault()` should be more like `[0]`

src/NetTopologySuite.Lab/Hull/ChiShape.cs
```diff
@@ -0,0 +1,151 @@
...
+            if (mergedLine.IsRing)
+            {
+                var lr = new LinearRing(mergedLine.CoordinateSequence, fact);
+
+                var concaveHull = new Polygon(lr, null, fact);
+
+                return concaveHull;
+            }
+            return mergedLine;
```
airbreather: Documentation says that this is supposed to throw if we can't generate a hull, but this returns a 1-dimensional geometry... which is the correct behavior?

src/NetTopologySuite.Lab/Hull/ChiShape.cs
airbreather: follow-up: "these two lines" was supposed to be a multi-line comment targeting 86 and 87... apparently multi-line comments are still in beta

src/NetTopologySuite.Lab/Hull/ChiShape.cs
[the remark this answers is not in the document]
grootstebozewolf: Not sure if the SortedList makes this O(n²), we should test this in a lab environment :)

src/NetTopologySuite.Lab/Hull/ChiShape.cs
grootstebozewolf: True

src/NetTopologySuite.Lab/Hull/ChiShape.cs
grootstebozewolf: Do we need to solve this edge case :)

grootstebozewolf: Can you help me with creating a test for this, not sure when this would be an issue.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
grootstebozewolf: Check!

src/NetTopologySuite.Lab/Hull/ChiShape.cs
grootstebozewolf: We will leave this as an exercise for the viewers.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
grootstebozewolf: I guess this is another edge case, when the geometry only has one or two points we should not throw something but return a defualt null value.

src/NetTopologySuite.Lab/Hull/ChiShape.cs
grootstebozewolf: Let's just throw an exception and call it a day.

PR #351, opened 2019-09-27
Fix xml code documentation flaws

FObermaier:


DGuidi:
not sure about latest commit you made...
`<value>` tag is used as return value in properties, at least according to [xml specs](https://docs.microsoft.com/it-it/dotnet/csharp/codedoc)

> The **value** tag is similar to the **returns** tag, except that you use it for properties

where `<c>` is used to mark part of the docs as **code**

FObermaier:
@DGuidi but `<value>` was used at other places, too. These were changed.

DGuidi:
My bad

DGuidi:
today it's definitively not my day...

Issue #366, opened 2019-11-20
STRtree<T>.Remove does not work when type parameter T is a value type


airbreather:
https://github.com/NetTopologySuite/NetTopologySuite/blob/c822013e89d879e2a52bcd1d0d3e095274645131/src/NetTopologySuite/Index/Strtree/AbstractSTRtree.cs#L371

This is equivalent to [what JTS does](https://github.com/locationtech/jts/blob/a2fc2b62bdcab7f92861a05018102cd64e1f3d79/modules/core/src/main/java/org/locationtech/jts/index/strtree/AbstractSTRtree.java#L385), but since ours is generic, we might receive an instance of a value-type, and so when the boxes are allocated for the `ReferenceEquals` call, they will (almost) always be two separate boxes, which means this will (almost) never work for instances of value types.

It's not a stretch to imagine the JIT reusing boxes here (even though it would technically violate ECMA), but ultimately we just flat-out shouldn't be calling `ReferenceEquals` with value-type instances.

My safest suggestion is:
```csharp
if (typeof(T).IsValueType)
{
    // use EqualityComparer<T>.Default.Equals
}
else
{
    // use ReferenceEquals
}
```

This should fix things that we **know** are broken, without breaking anything reasonable.

In an ideal world, however, we ought to always use `EqualityComparer<T>.Default.Equals` and/or `Object.Equals` on this line (and/or allow injecting an `IEqualityComparer<T>` to use), but:
1. JTS uses strict reference equality, so this would be a deviation.
1. Existing code might use a type that has an overridden `Equals(object obj)` and/or that might implement `IEquatable<T>` in a way that they don't expect to work with this... I think it's not a **huge** deal to break that stuff, since `List<T>.Remove` and friends work the same way, but there it is.

airbreather:
Reported as locationtech/jts#497

FObermaier:
Can this be closed?

airbreather:
I think I'd like to keep it open unless we decide for sure that we're not going to do anything about it until JTS does.

`STRtree<T>.Remove` and `Quadtree<T>.Remove` look like they would behave differently, and I think it would be weird for someone to rely on this behavior.

PR #374, opened 2019-12-10
Allow STRtree<T>.Remove to work with value types

airbreather:
Workaround for #366

I think the **ideal** would be to allow the caller to inject the `IEqualityComparer<TItem>` (or something equivalent), or else to **always** use the default `IEqualityComparer<TItem>` for the type.

This approach, however, ensures that the only observable changes will be in cases that would have **never** worked before (or at least, they would **only** have worked on runtimes that do the same kind of box reuse that the JVM exhibits in the repro I made for locationtech/jts#497, which technically violates ECMA-355 §III.4.1 even though nobody in their right mind ought to care about that particular violation).

This approach also gives us the most flexibility to port whatever locationtech/jts#497 turns into down the road: there's no API change, and the only observable behavior change should be a subset of the changes that would come from any realistic fix for that issue.

### DIFF
--- a/src/NetTopologySuite/Index/Strtree/AbstractSTRtree.cs
+++ b/src/NetTopologySuite/Index/Strtree/AbstractSTRtree.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using NetTopologySuite.Geometries;
 using NetTopologySuite.Utilities;
 using IList = System.Collections.Generic.IList<object>;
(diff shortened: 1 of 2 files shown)


src/NetTopologySuite/Index/Strtree/AbstractSTRtree.cs
```diff
@@ -368,8 +371,16 @@ private static bool RemoveItem(AbstractNode<T, TItem> node, TItem item)
             for (var i = node.ChildBoundables.GetEnumerator(); i.MoveNext(); )
             {
                 var childBoundable = i.Current as ItemBoundable<T, TItem>;
-                if (childBoundable != null && ReferenceEquals(childBoundable.Item, item))
+                if (childBoundable != null)
+                {
+                    // EqualityComparerForRemoveItem might be null, to signal that we should use the
+                    // default equality comparer for the type.
+                    if (EqualityComparerForRemoveItem?.Equals(childBoundable.Item, item) ??
+                        EqualityComparer<TItem>.Default.Equals(childBoundable.Item, item))
```
FObermaier: Are you sure that ??-ing to the default comparer for value types for **each** `childBoundable` is actually faster than setting default comparer within `InitializeEqualityComparerForRemoveItem`?

airbreather: Yes, but it probably doesn't matter here.  I'll replace this with the more obvious version.

PR #392, opened 2020-03-17
Fix Linux builds.

airbreather:
We can use `Microsoft.DotNet.ApiCompat` with less setup.
Also remove the "This package references GeoAPI." line from the package description.


airbreather:
Note that builds are failing, but for reasons other than the things in here.

FObermaier:
I've been looking for `PackageDownload` task/target.

PR #412, opened 2020-05-27
Enhancement/overlay-sr

FObermaier:
Overlay with snap-rounding

* https://lin-ear-th-inking.blogspot.com/2020/05/jts-overlay-next-generation.html
* https://github.com/dr-jts/jts/tree/overlay-sr


FObermaier:
**Note: Work in progress**

airbreather:
I'll look into the API later... having the algorithm do its own snap-rounding is going to be a **very** welcome change, so I'm excited to see this.

dr-jts:
A couple of comments on this port:

* The OverlayNG codebase is still getting a few bug fixes and refactorings, so make sure to track the commits on that branch.
* A recent addition is the [SnappingNoder](https://github.com/dr-jts/jts/blob/overlay-sr/modules/core/src/main/java/org/locationtech/jts/noding/snap/SnappingNoder.java), which provides much better robustness for floating precision overlay operations.  This has enabled the [OverlayNGSnapIfNeeded](https://github.com/dr-jts/jts/blob/overlay-sr/modules/core/src/main/java/org/locationtech/jts/operation/overlayng/OverlayNGSnapIfNeeded.java) class, which uses some heuristics to maximise performance and robustness.  This is the recommended class to use for floating-precision overlay.

FObermaier:
@dr-jts, I'm aware of that and I'm keeping an eye on it. But ATM I'm lacking time.

juliusfriedman:
@dr-jts, Glad to see these improvements.

1 thing might I suggest is would be the ability to limit Operations, e.g. If I only need 1 intersection then I should be able to improve the speed of the IntersectionOp by giving a max number of intersections to find, it would also then be trivial to resume from that index if more are needed.

FObermaier:
@juliusfriedman, why would you want only one intersection? If you need a predicate check possibly one against many I strongly urge you to use the functionality provided in the `NetTopologySuite.Geometries.Prepared` namespace.

juliusfriedman:
> @juliusfriedman, why would you want only one intersection? If you need a predicate check possibly one against many I strongly urge you to use the functionality provided in the `NetTopologySuite.Geometries.Prepared` namespace.

Thank you,

What also about if I want to buffer the 1st clockwise intersection a different way than I want to buffer the first counter clockwise intersection?

I can imagine a stepped algorithm which can step the intersections required and pass them to the buffer op or any other op in the chain where required for the processing required by the application...

Is that something which Prepared can also handle?

FObermaier:
> Is that something which Prepared can also handle?

No, that namespace is for predicates only.

FObermaier:
This is now in sync with the original.
What we need to settle:
* [ ] JTS' `OverlayLabeller` uses `Deque<OverlayEdge>`. There is no interface or class for a deque in .NetStandard. Currently a `LinkedList` is used as replacement, we need to consider pulling in our own IDeque<T> interface and inplementation.
* [ ] JTS' has quite some breaking changes which I avoided by attributing with `Obsolete` and remove internal usage. An example is the move from `Position` and `Quadrant` to geometries namespace.

What remains to be done:
* [ ] Improve xml-test reder to be able to parse and perform tests on newer functionality. I have a strong feeling that this is not working ATM.

FObermaier:
@airbreather, this PR is in the same state as the code merged to master in JTS.
The failing tests are due to missing functionality in our `NetTopologySuite.TestRunner` project.

airbreather:
> @airbreather, this PR is in the same state as the code merged to master in JTS.
> The failing tests are due to missing functionality in our `NetTopologySuite.TestRunner` project and old Xml-test-files.

OK thanks for the heads-up.  I'll do a review of the API, but I won't be able to do a full review of the code... there's a lot there.

Sorry I didn't mention this, but I did want to hold off until it got stable enough to get merged into JTS, since it was still in flux when I made my previous comment on the thread.

FObermaier:
Thanks, develop would be ready to merge, too.

I'd like to change `NetTopologySuite.Operation.OverlayNg` namespace to `NetTopologySuite.Operation.OverlayNG` prior to merging.

airbreather:
re: `Deque<OverlayEdge>`, JTS only calls `isEmpty()`, `addFirst()`, and `removeFirst()`, so we can actually use a `Stack<OverlayEdge>`.

dr-jts:
> re: `Deque<OverlayEdge>`, JTS only calls `isEmpty()`, `addFirst()`, and `removeFirst()`, so we can actually use a `Stack<OverlayEdge>`.

Yes, the `Deque` is being used as a stack (as is apparently Java best practice now).

src/NetTopologySuite/Operation/OverlayNG/PrecisionUtility.cs
```diff
@@ -0,0 +1,350 @@
...
+        /// NOTE: this is a heuristic determination, so is not guaranteed to
+        /// eliminate precision issues.
+        /// <para/>
+        /// WARNING: this is <b>very</b> slow.
+        /// </summary>
+        /// <param name="a">A geometry</param>
+        /// <returns>A suitable precision model for overlay</returns>
+        public static PrecisionModel RobustPM(Geometry a)
+        {
+            double scale = PrecisionUtility.RobustScale(a);
```
airbreather, line 296: validate `a` non-null before using it

(we could rely on `RobustScale` to validate for us if we do an earlier comment, mainly just mentioning this here for consistency)

src/NetTopologySuite/Operation/OverlayNG/PrecisionUtility.cs
```diff
@@ -0,0 +1,350 @@
...
+        /// inherent scale and the safe scale factors.
+        /// <para/>
+        /// WARNING: this is <b>very</b> slow.
+        /// </summary>
+        /// <param name="a">A geometry</param>
+        /// <param name="b">A geometry</param>
+        /// <returns>A scale factor for use in overlay operations</returns>
+        public static double RobustScale(Geometry a, Geometry b)
+        {
+            double inherentScale = InherentScale(a, b);
```
airbreather, line 314: validate `a` non-null before using it

(we could rely on `InherentScale` to validate for us if we do an earlier comment, mainly just mentioning this here for consistency)

src/NetTopologySuite/Operation/OverlayNG/PrecisionUtility.cs
```diff
@@ -0,0 +1,350 @@
...
+        /// the digits of precision and is
+        /// safe to use for overlay operations.
+        /// The robust scale is the minimum of the
+        /// inherent scale and the safe scale factors.
+        /// </summary>
+        /// <param name="a">A geometry</param>
+        /// <returns>A scale factor for use in overlay operations</returns>
+        public static double RobustScale(Geometry a)
+        {
+            double inherentScale = InherentScale(a);
```
airbreather, line 330: validate `a` non-null before using it

(we could rely on `InherentScale` to validate for us if we do an earlier comment, mainly just mentioning this here for consistency)

src/NetTopologySuite/Operation/OverlayNG/RingClipper.cs
```diff
@@ -0,0 +1,192 @@
...
+    /// The clipping envelope should be generated using <see cref="RobustClipEnvelopeComputer"/>,
+    /// to ensure that intersecting line segments are not perturbed
+    /// by clipping.
+    /// This is required to ensure that the overlay of the
+    /// clipped geometry is robust and correct (i.e. the same as
+    /// if clipping was not used).
+    /// </summary>
+    /// <seealso cref="LineLimiter"/>
+    /// <author>Martin Davis</author>
+    public class RingClipper
```
airbreather, line 30: this isn't designed to be inherited from, so let's mark it `sealed`

(We can always unseal it later if that becomes a problem)

src/NetTopologySuite/Operation/OverlayNG/RingClipper.cs
```diff
@@ -0,0 +1,192 @@
...
+        private readonly double _clipEnvMinX;
+        private readonly double _clipEnvMaxX;
+
+        /// <summary>
+        /// Creates a new clipper for the given envelope.
+        /// </summary>
+        /// <param name="clipEnv">The clipping envelope</param>
+        public RingClipper(Envelope clipEnv)
+        {
+            //this.clipEnv = clipEnv;
```
airbreather, line 50: validate `clipEnv` non-null before using it

src/NetTopologySuite/Operation/OverlayNG/RingClipper.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        /// <summary>
+        /// Clips a list of points to the clipping rectangle box.
+        /// </summary>
+        /// <param name="pts">The points of the ring</param>
+        /// <returns>The points of the clipped ring</returns>
+        public Coordinate[] Clip(Coordinate[] pts)
+        {
+            for (int edgeIndex = 0; edgeIndex < 4; edgeIndex++)
```
airbreather, line 64: validate `pts` non-null before using it

src/NetTopologySuite/Operation/OverlayNG/RobustClipEnvelopeComputer.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <summary>
+    /// Computes a robust clipping envelope for a pair of polygonal geometries.
+    /// The envelope is computed to be large enough to include the full
+    /// length of all geometry line segments which intersect
+    /// a given target envelope.
+    /// This ensures that line segments which might intersect are
+    /// not perturbed when clipped using <see cref="RingClipper"/>.
+    /// </summary>
+    /// <author>Martin Davis</author>
+    internal class RobustClipEnvelopeComputer
```
airbreather, line 14: this isn't designed to be inherited from, so let's mark it `sealed`

(We can always unseal it later if that becomes a problem)

src/NetTopologySuite/Operation/OverlayNG/UnaryUnionNG.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace NetTopologySuite.Operation.OverlayNg
+{
+    /// <summary>
+    /// Unions a collection of geometries in an
+    /// efficient way, using {@link OverlayNG}
+    /// to ensure robust computation.
+    /// </summary>
+    /// <author>Martin Davis</author>
+    public class UnaryUnionNG
```
airbreather, line 13: Instead of a private constructor, let's make this `static`.

src/NetTopologySuite/Operation/OverlayNG/UnaryUnionNG.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Unions a collection of geometries
+        /// using a given precision model.
+        /// </summary>
+        /// <param name="geom">The geometry to union</param>
+        /// <param name="pm">The precision model to use</param>
+        /// <returns>The union of the geometries</returns>
+        public static Geometry Union(Geometry geom, PrecisionModel pm)
+        {
+            var unionSRFun = new UnionStrategy((g0, g1) =>
```
airbreather, line 24: validate `geom` non-null before using it

src/NetTopologySuite/Operation/Union/CascadedPolygonUnion.cs
```diff
@@ -42,11 +46,41 @@ namespace NetTopologySuite.Operation.Union
...
+            {
+                try
+                {
+                    return SnapIfNeededOverlayOp.Union(g0, g1);
+                }
+                catch (TopologyException ex)
+                {
+                    // union-by-buffer only works for polygons
+                    if (g0.Dimension != Dimension.Surface || g1.Dimension != Dimension.Surface)
+                        throw ex;
```
airbreather, line 64: instead of this conditional rethrow (which should be `throw;` instead of `throw ex;`), use a filtered catch:

```csharp
catch (TopologyException) when (g0.Dimension == Dimension.Surface && g1.Dimension == Dimension.Surface)
{
    return UnionPolygonsByBuffer(g0, g1);
}
```

src/NetTopologySuite/Operation/Union/CascadedPolygonUnion.cs
```diff
@@ -42,11 +46,41 @@ namespace NetTopologySuite.Operation.Union
     /// <seealso href="http://code.google.com/p/nettopologysuite/issues/detail?id=44"/>
     public class CascadedPolygonUnion
     {
-        private ICollection<Geometry> _inputPolys;
-        private GeometryFactory _geomFactory;
-#if UseWorker
-        private int _numThreadsStarted = 0;
-#endif
+        /// <summary>
+        /// A union strategy that uses the classic NTS {@link SnapIfNeededOverlayOp},
```
airbreather, line 50: `@link` should be `<see cref=`

src/NetTopologySuite/Operation/Union/CascadedPolygonUnion.cs
```diff
@@ -59,15 +93,47 @@ public static Geometry Union(ICollection<Geometry> polys)
...
+            : this(polys, ClassicUnion)
+        {
+        }
+
+        /**
+      * Creates a new instance to union
+      * the given collection of {@link Geometry}s.
+      *
+      * @param polys a collection of {@link Polygonal} {@link Geometry}s
+      */
```
airbreather, line 127: convert javadoc to xmldoc

src/NetTopologySuite/Operation/Union/CascadedPolygonUnion.cs
```diff
@@ -59,15 +93,47 @@ public static Geometry Union(ICollection<Geometry> polys)
...
+            var op = new CascadedPolygonUnion(polys, unionStrategy);
+            return op.Union();
+        }
+
+        private ICollection<Geometry> _inputPolys;
+        private GeometryFactory _geomFactory;
+        private readonly UnionStrategy _unionStrategy;
+
+        private int countRemainder = 0;
+        private int countInput = 0;
```
airbreather, line 107: can we prefix field names with `_`

src/NetTopologySuite/Operation/Union/OverlapUnion.cs
```diff
@@ -65,21 +68,48 @@ public static Geometry Union(Geometry g0, Geometry g1)
...
+        /// Union a pair of geometries,
+        /// using the more performant overlap union algorithm if possible.
+        /// </summary>
+        /// <param name="g0">A geometry to union</param>
+        /// <param name="g1">A geometry to union</param>
+        /// <param name="unionFun">Function to union two geometries</param>
+        /// <returns>The union of the inputs</returns>
+        public static Geometry Union(Geometry g0, Geometry g1, UnionStrategy unionFun)
+        {
+            var union = new OverlapUnion(g0, g1, unionFun);
```
airbreather, line 81: validate `g0` non-null before using it

src/NetTopologySuite/Operation/Union/UnionStrategy.cs
```diff
@@ -0,0 +1,43 @@
...
+using NetTopologySuite.Geometries;
+
+namespace NetTopologySuite.Operation.Union
+{
+    /// <summary>
+    /// An strategy class that allows UnaryUnion to adapt to different
+    /// kinds of overlay algorithms.
+    /// </summary>
+    /// <author>Martin Davis</author>
+    public class UnionStrategy
```
airbreather, line 11: this isn't designed to be inherited from, so let's mark it `sealed`

(We can always unseal it later if that becomes a problem)

src/NetTopologySuite/Operation/Union/UnionStrategy.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Computes the union of two geometries.
+        /// This method may throw a <see cref="TopologyException"/>
+        /// if one is encountered.
+        /// </summary>
+        /// <param name="g0">A geometry</param>
+        /// <param name="g1">A geometry</param>
+        /// <returns>The union of the input</returns>
+        public Geometry Union(Geometry g0, Geometry g1) => _unionFunction(g0, g1);
```
airbreather, line 29: validate `g0` non-null before using it

also `g1`???

or just make this `internal`???

src/NetTopologySuite/Planargraph/DirectedEdge.cs
```diff
@@ -85,7 +85,14 @@ public Edge Edge
...
         /// </summary>
-        public int Quadrant => _quadrant;
+        [Obsolete("Use QuadrantValue")]
+        public int Quadrant => (int)_quadrant;
+
+        /// <summary>
+        /// Returns 0, 1, 2, or 3, indicating the quadrant in which this DirectedEdge's
+        /// orientation lies.
+        /// </summary>
+        public Quadrant QuadrantValue => _quadrant;
```
airbreather, line 95: Not sure about this name... maybe [`QuadrantEx`](https://stackoverflow.com/q/4039054/1083771)?

airbreather (changes requested):
I only looked at the non-test files.

Let me know if you want me to fix all of these items.  I don't mind.

src/NetTopologySuite/Geometries/Position.cs
```diff
@@ -0,0 +1,55 @@
+namespace NetTopologySuite.Geometries
+{
+    /// <summary>
+    /// Indicates the position of a location relative to a
+    /// node or edge component of a planar topological structure.
+    /// </summary>
+    public enum Position
```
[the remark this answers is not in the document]
FObermaier: What I like about enums is that I don't just see the value but also its meaning.   
Maybe we can mix both, rename `Position` enum to `Position[Index|Value]` and then use that inside Position enum-class.
What is the benefit of an enum-class over a struct in this case?

src/NetTopologySuite/Geometries/Quadrant.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// The quadants are numbered as follows:
+    /// <para>
+    /// <code>
+    /// 1 - NW | 0 - NE
+    /// -------+-------
+    /// 2 - SW | 3 - SE
+    /// </code>
+    /// </para>
+    /// </remarks>
+    public enum Quadrant
```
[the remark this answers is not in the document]
FObermaier: dito Position.

src/NetTopologySuite/GeometriesGraph/Depth.cs
```diff
@@ -95,6 +126,7 @@ public Location GetLocation(int geomIndex, Positions posIndex)
         /// <param name="geomIndex"></param>
         /// <param name="posIndex"></param>
         /// <param name="_location"></param>
+        [Obsolete]
```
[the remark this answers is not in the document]
FObermaier: It is not used throughout the code-base. Neither in JTS nor in NTS.

src/NetTopologySuite/Operation/Union/CascadedPolygonUnion.cs
FObermaier: I did not touch that file, did I?

src/NetTopologySuite/Planargraph/DirectedEdge.cs
FObermaier: Should be in sync with how we name it in `EdgeEnd`

FObermaier:
> Let me know if you want me to fix all of these items. I don't mind.

That'd be great

src/NetTopologySuite/Geometries/Position.cs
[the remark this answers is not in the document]
airbreather: >  What is the benefit of an enum-class over a struct in this case?

It could be a struct instead, the point I wanted to make is just that a .NET `enum` can't have methods or properties, so in trying to port one Java class this way, we wind up making two .NET types, with one of them feeling out of place.

In most places, I also favor designing around .NET `enum`s, but obviously our design is more constrained in core NTS.

src/NetTopologySuite/GeometriesGraph/Depth.cs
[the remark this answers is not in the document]
airbreather: ok, just updated the message then

src/NetTopologySuite/GeometriesGraph/QuadrantOp.cs
```diff
@@ -34,6 +35,7 @@ public class QuadrantOp
         /// <summary>
         /// Only static methods!
         /// </summary>
+        [Obsolete]
```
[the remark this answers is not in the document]
airbreather: removed the private constructor altogether and just made the class `static`

src/NetTopologySuite/GeometriesGraph/QuadrantOp.cs
```diff
@@ -11,8 +11,8 @@ namespace NetTopologySuite.GeometriesGraph
     /// 2 | 3
     /// </para>
     /// </summary>
-    [Obsolete("Use NetTopologySuite.Topology.Quadrant and NetTopologySuite.Topology.QuadrantExtensions")]
-    public class QuadrantOp
+    [Obsolete("Use NetTopologySuite.Geometries.Quadrant and NetTopologySuite.Geometries.QuadrantExtensions")]
+    public static class QuadrantOp
```
FObermaier, line 15: This is not a breaking change?

airbreather: It shouldn't be -- since `QuadrantOp` can't be instantiated, the only thing you can do with it is call static methods.

I guess it's theoretically possible that someone's doing something like this:
```csharp
QuadrantOp op = null;
```

but that's really far-fetched.

airbreather: ahh, the API Compat check actually rejects it, so I'm going to revert.

src/NetTopologySuite/Operation/OverlayNG/Edge.cs
```diff
@@ -0,0 +1,336 @@
...
+
+        public Coordinate[] Coordinates
+        {
+            get => _pts;
+        }
+
+        public Coordinate GetCoordinate(int index)
+        {
+            return _pts[index];
+        }
```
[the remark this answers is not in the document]
airbreather: on second thought, no

src/NetTopologySuite/Operation/OverlayNG/PrecisionUtility.cs
```diff
@@ -0,0 +1,350 @@
...
+            }
+
+            private void UpdateScaleMax(double value)
+            {
+                double scaleVal = PrecisionUtility.InherentScale(value);
+                if (scaleVal > _scale)
+                {
+                    //System.out.println("Value " + value + " has scale: " + scaleVal);
+                    _scale = scaleVal;
+                }
```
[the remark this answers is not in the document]
airbreather: This approach is respectably fast, but there are values that cause it to give incorrect answers (probably dotnet/runtime#1643).

The following returns the same value as what's there, in all the cases I've tested, plus those other cases that failed the above, and it's even faster (for real-world values) than the broken `Math.Round` alternative:
```csharp
public static int NumberOfDecimals(double value)
{
    if (value != Math.Round(value, 15) || value > 79228162514264337593543950335d || value < -79228162514264337593543950335d)
    {
        return NumberOfDecimalsRare(value);
    }

    decimal valueAsDecimal = (decimal)value;
    return ((byte*)&valueAsDecimal)[2];
}
```

src/NetTopologySuite/Operation/Union/UnionStrategy.cs
airbreather: just making it `internal`.  it still builds.

airbreather:
> * [ ] Improve xml-test reader to be able to parse and perform tests on newer functionality. I have a strong feeling that this is not working ATM.
>
> The failing tests are due to missing functionality in our `NetTopologySuite.TestRunner` project and old Xml-test-files.

The only issue that I directly noticed here was that we were checking for the old `com.vividsolutions` namespaces in a switch statement where the `default` branch had a `Console.ReadKey` call, so those tests would never terminate.  After fixing that, everything passes (though, as I mentioned, I haven't looked at the code for any of the tests).

There's one more failing test: `Fixing_invalid_polygon_with_Buffer_0_returns_empty_polygon`.  I marked it as `[Category("FailureCase")]` for now, but **just** to make sure that everything else gets a green checkmark from Travis-CI (it does :grinning:).

I left the "enum class" feedback open on `Position` and `Quadrant`, since I'm not sure what you want to do about it.  I personally favor that kind of approach since it seems to give the best of both worlds: there's just one type like JTS has (`Position` / `Quadrant`), and yet we still get to use strong typing so that the methods don't just accept any old `int` like they do in JTS.
- If you're OK with me making the change, let me know and I will.

airbreather:
image

(sorry, mis-clicked)

src/NetTopologySuite/Geometries/Position.cs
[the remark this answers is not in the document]
FObermaier: I refactored `Position` to a struct.

src/NetTopologySuite/Geometries/Quadrant.cs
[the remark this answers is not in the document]
FObermaier: I refactored `Quadrant` to a struct.

test/NetTopologySuite.Samples.Console/Tests/Github/Issues.cs
```diff
@@ -235,7 +235,8 @@ public void LineString_Intersection_with_Polygon_return_null1()
             Debug.WriteLine(res.AsText());
         }
 
-        [Test(Description = "GitHub Issue #125")]
+        [Test(Description = "GitHub Issue #124")]
+        [Category("FailureCase")]
```
airbreather, line 239: We should still figure out why this test fails... is it a bug upstream?

airbreather (approved):
I'm fine if you want to merge this huge branch before we figure out that test failure, but we should probably figure it out before the next release.

test/NetTopologySuite.Samples.Console/Tests/Github/Issues.cs
dr-jts: This is because the polygon is actually a figure-8, and buffer(0) often doesn't handle those correctly.

This test probably isn't relevant any more, because `GeometryPrecisionReducer` now uses Snap-Rounding, which does not use buffer(0) to creating valid topology.

airbreather: @dr-jts the exception is actually thrown in `GeometryPrecisionReducer` itself, specifically [here](https://github.com/NetTopologySuite/NetTopologySuite/blob/bead06f69f2cfab88dde7df96d2b4b351abeeaf4/src/NetTopologySuite/Operation/OverlayNG/PolygonBuilder.cs#L190).  I haven't tried this in JTS to determine whether this is a problem with the algorithm or with our port.

⌛ if I get the time to get around to it, I'll try it out and report it in JTS if it fails.

dr-jts: Ah, I understand now.  JTS throws the same error.  And in fact it throws for any precision value.  The problem is that the input geometry is invalid.  SInce `GeometryPrecisionReducer` now uses OverlayNG, it is more sensitive to invalidity.  This might be considered a regression, I suppose - but it still seems like a reasonable & consistent contract that input needs to be valid.  

The usual heuristic to fix bad geometry is to use `buffer(0)`, but in this case as noted that fails, producing a `POLYGON EMPTY`.  It is only a heuristic, and never guaranteed to work.  At some point hopefully JTS will acquire a `MakeValid` function.

dr-jts: The cause of the invalidity is a "bowtie" self-intersection at the right end of the polygon.  The points forming it are only on the order of 1e-15 apart!

image

It is interesting that this fails, since you would expect these points to snap together.  This might actual reveal a flaw in the OverlayNG algorithm, in that it uses the original polygon to check for containment, which leaves it susceptible to input invalidity.  I should probably revisit this design.

airbreather:
:tada:

DGuidi:
👏👏👏

test/NetTopologySuite.Samples.Console/Tests/Github/Issues.cs
dr-jts: The reason this case fails in `GeometryPrecisionReducer` is because of the invalid input.  Because of the bowtie polygon, the orientation of the ring is ambiguous.  It is evaluated in such a way that the large portion of the ring has orientation indicating that it is a hole.  The smaller "shell" part collapses to a single point, leaving just a hole ring.  This then of course cannot be assigned to a shell.

The throwing of a `TopologyException` is actually a good thing, since it indicates there is something wrong with the input.   It's a bit more worrying that this polygon can actually produce output when used in say intersections (although the output is invalid).  At the moment I don't have an idea of how to improve this.

PR #425, opened 2020-07-15
Implemented ConcaveHull

ststeiger:
### Prerequisites

- [ x ] I have written a descriptive pull-request title
- [ x ] I have verified that there are no overlapping [pull-requests](https://github.com/NetTopologySuite/NetTopologySuite/pulls) open
- [ x ] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository.

- [ x ] I have provided test coverage for my change (where applicable)

### Description
Implemented ConcaveHull by converting the Java Source of [Concave_Hull ](http://www.rotefabrik.free.fr/concave_hull/) to C#. 
Of the existing source, the use of DelaunayTriangulationBuilder was also wrong. 
Correct is ConformingDelaunayTriangulationBuilder.

### DIFF
--- a/src/NetTopologySuite/Triangulate/QuadEdge/Vertex.cs
+++ b/src/NetTopologySuite/Triangulate/QuadEdge/Vertex.cs
@@ -30,8 +30,40 @@ namespace NetTopologySuite.Triangulate.QuadEdge
         private const int ORIGIN = 5;
         private const int DESTINATION = 6;
 
+
+        
+        private int id; // ID of the vertex	
         private readonly Coordinate _p;
+        private bool border;
         // private int edgeNumber = -1;
+
+        /// <summary>	
+        /// Default constructor	
+        /// </summary>	
+        public Vertex()
+        { }
+
+        /// <summary>	
+        /// Creates an instance of this class using the given id	
+        /// </summary>	
+        /// <param name="id">ID of the vertex</param>	
+        public Vertex(int id)
+        {
+            this.id = id;
+        }
+
+
+        /// <summary>	
+        /// Creates an instance of this class using the given id	
+        /// </summary>	
+        /// <param name="id">ID of the vertex</param>	
+        /// <param name="border">defines if the vertex is a border vertex or not in the triangulation framework</param>	
+        public Vertex(int id, bool border)
+        {
+            this.id = id;
+            this.border = border;
+        }
+
 
         /// <summary>
         /// Creates an instance of this class using the given x- and y-ordinate valuse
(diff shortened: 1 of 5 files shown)


ststeiger:
Signed CLA.

src/NetTopologySuite/Triangulate/QuadEdge/Vertex.cs
```diff
@@ -30,9 +30,41 @@ public class Vertex : IEquatable<Vertex>
         private const int ORIGIN = 5;
         private const int DESTINATION = 6;
 
+
+        
+        private int id; // ID of the vertex	
         private readonly Coordinate _p;
+        private bool border;
```
airbreather, line 37: Per the summary documentation on this class, the intent is to make your own subclass to add these properties, instead of forcing everyone to pay for them.

Unfortunately there's no convenient way to do this using `ConformingDelaunayTriangulationBuilder`, you have to use `ConformingDelaunayTriangulator` more directly.

airbreather (changes requested):
# Non-technical

> Implemented ConcaveHull by converting the Java Source of [Concave_Hull ](http://www.rotefabrik.free.fr/concave_hull/) to C#.

That makes this a bit complicated: the source code from that post is LGPL, so unless you have special permission to send this to us under a different license, then all it can do is (at best) stay in NetTopologySuite.Lab or somewhere else that's separate from the BSD-3-Clause licensed code...

# Technical

I didn't go through the actual lab code, since it doesn't get distributed anywhere.  See comment on `Vertex`.

src/NetTopologySuite/Triangulate/QuadEdge/Vertex.cs
```diff
@@ -30,9 +30,41 @@ public class Vertex : IEquatable<Vertex>
...
+        /// <summary>	
+        /// Creates an instance of this class using the given id	
+        /// </summary>	
+        /// <param name="id">ID of the vertex</param>	
+        /// <param name="border">defines if the vertex is a border vertex or not in the triangulation framework</param>	
+        public Vertex(int id, bool border)
+        {
+            this.id = id;
+            this.border = border;
+        }
```
BrutalHex, line 65: "defines if the vertex is a border vertex or not"
we should have two classes: BorderVertex and Vertex . since we gonna have multiple types of vertices thus addition of an abstract class as a base for  BorderVertex and Vertex would be better.

FObermaier:
NTS now has a ConcaveHull implementation:
https://github.com/NetTopologySuite/NetTopologySuite/blob/develop/src/NetTopologySuite/Algorithm/Hull/ConcaveHull.cs

Issue #447, opened 2020-11-11
Wrong equals operation for points leads to wrong save behavior in EntityFramework


Danielku15:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of NetTopologySuite
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/NetTopologySuite/NetTopologySuite/issues) and [closed](https://github.com/NetTopologySuite/NetTopologySuite/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
NTS has a design issue that the equality check of 2 points is only done on 2D base. We are having the issue that we are updating some coordinates to add the height component if it is missing. But due to the fact that `new Point(1,2) == new Point(1,2,3)` (including `Equals` and `EqualsExact`), Entity Framework will not detect any change in the value and will not generate an update statement.

![image](https://user-images.githubusercontent.com/674916/98841901-0d57c300-2449-11eb-9fca-cdb7a215545b.png)

### Steps to Reproduce

```csharp
Assert.IsFalse(new NetTopologySuite.Geometries.Point(1,2) == new NetTopologySuite.Geometries.Point(1,2,3))
Assert.AreNotEqual(new NetTopologySuite.Geometries.Point(1,2), new NetTopologySuite.Geometries.Point(1,2,3))
```

### System Configuration
Tell us about the environment where you are experiencing the bug:
- NetTopologySuite version: 2.0.0
- Other NetTopologySuite packages and versions: 
  - Microsoft.EntityFrameworkCore.SqlServer.NetTopologySuite 3.1.6
  - NetTopologySuite.IO.SqlServerBytes 2.0.0
- Environment (Operating system, version and so on): Windows 10 Enterprise 2004
- .NET Framework version: .net 4.8
- Additional information: 

Here we can see that the equals operation is only respecting 2D by design: 

https://github.com/NetTopologySuite/NetTopologySuite/blob/7babd53a14f452ec12aee6c15a6f5f943d26aa5d/src/NetTopologySuite/Geometries/Coordinate.cs#L259-L262

NTS should correctly respect the availability of the height component and consider it for equality checks. Especially in regards to EntityFramework where NTS is also used for Spatial Operations it is very essential to respect also the height component. Even more when working in the geodetic domain.

DGuidi:
[...]

PR #452, opened 2020-11-27
Control Geometry overlay algorithm via factory setting

FObermaier:
With https://github.com/locationtech/jts/commit/31cb7b640a8b9552b912a6ac96a6e5795b6c0957 JTS delegated computation of  Geometry overlay functions to a special class (`GeometryOverlay`). It controls which algorithm to use by evaluating a system property.

In regard to #447 and the multiple compaints abount distance and area computations in EF context we should consider to take another approach and delegate these operations to the GeometryFactory.

### DIFF
--- a/src/NetTopologySuite/Geometries/Geometry.cs
+++ b/src/NetTopologySuite/Geometries/Geometry.cs
@@ -2,18 +2,14 @@ using System;
 using System.Collections.Generic;
 using System.Xml;
 using NetTopologySuite.Algorithm;
-using NetTopologySuite.Geometries.Utilities;
 using NetTopologySuite.IO;
 using NetTopologySuite.IO.GML2;
 using NetTopologySuite.Operation;
 using NetTopologySuite.Operation.Buffer;
 using NetTopologySuite.Operation.Distance;
 using NetTopologySuite.Operation.Linemerge;
-using NetTopologySuite.Operation.Overlay;
-using NetTopologySuite.Operation.Overlay.Snap;
 using NetTopologySuite.Operation.Predicate;
 using NetTopologySuite.Operation.Relate;
-using NetTopologySuite.Operation.Union;
 using NetTopologySuite.Operation.Valid;
 
 namespace NetTopologySuite.Geometries
@@ -1515,8 +1511,11 @@ namespace NetTopologySuite.Geometries
         /// <returns>A geometry representing the point-set common to the two <c>Geometry</c>s.</returns>
         /// <exception cref="TopologyException">if a robustness error occurs.</exception>
         /// <exception cref="ArgumentException">if the argument is a non-empty heterogeneous <c>GeometryCollection</c></exception>
+        /// <exception cref="ArgumentException">if the argument has a factory with a different <c>GeometryOverlay</c> object assigned</exception>
         public Geometry Intersection(Geometry other)
         {
+            if (!Equals(other.Factory.GeometryOverlay, Factory.GeometryOverlay))
+                throw new ArgumentException("Geometry has factory with different GeometryOverlay operation assigned", nameof(other));
             return Factory.GeometryOverlay.Intersection(this, other);
         }
 
@@ -1552,9 +1551,12 @@ namespace NetTopologySuite.Geometries
         /// points of <c>other</c></returns>
         /// <exception cref="TopologyException">Thrown if a robustness error occurs</exception>
         /// <exception cref="ArgumentException">Thrown if either input is a non-empty GeometryCollection</exception>
+        /// <exception cref="ArgumentException">if the argument has a factory with a different <c>GeometryOverlay</c> object assigned</exception>
         /// <seealso cref="LineMerger"/>
         public Geometry Union(Geometry other)
         {
+            if (!Equals(other.Factory.GeometryOverlay, Factory.GeometryOverlay))
+                throw new ArgumentException("Geometry has factory with different GeometryOverlay operation assigned", nameof(other));
             return Factory.GeometryOverlay.Union(this, other);
         }
 
@@ -1570,8 +1572,11 @@ namespace NetTopologySuite.Geometries
         /// </summary>
         /// <param name="other">The <c>Geometry</c> with which to compute the difference.</param>
         /// <returns>A Geometry representing the point-set difference of this <c>Geometry</c> with <c>other</c>.</returns>
+        /// <exception cref="ArgumentException">if the argument has a factory with a different <c>GeometryOverlay</c> object assigned</exception>
         public Geometry Difference(Geometry other)
         {
+            if (!Equals(other.Factory.GeometryOverlay, Factory.GeometryOverlay))
+                throw new ArgumentException("Geometry has factory with different GeometryOverlay operation assigned", nameof(other));
             return Factory.GeometryOverlay.Difference(this, other);
         }
 
@@ -1588,8 +1593,11 @@ namespace NetTopologySuite.Geometries
         /// </summary>
         /// <param name="other">The <c>Geometry</c> with which to compute the symmetric difference.</param>
         /// <returns>a Geometry representing the point-set symmetric difference of this <c>Geometry</c> with <c>other</c>.</returns>
+        /// <exception cref="ArgumentException">if the argument has a factory with a different <c>GeometryOverlay</c> object assigned</exception>
         public Geometry SymmetricDifference(Geometry other)
         {
+            if (!Equals(other.Factory.GeometryOverlay, Factory.GeometryOverlay))
+                throw new ArgumentException("Geometry has factory with different GeometryOverlay operation assigned", nameof(other));
             return Factory.GeometryOverlay.SymmetricDifference(this, other);
         }
 
--- a/src/NetTopologySuite/Geometries/Geometry.cs
+++ b/src/NetTopologySuite/Geometries/Geometry.cs
@@ -354,7 +354,8 @@ namespace NetTopologySuite.Geometries
         /// In general, the array cannot be assumed to be the actual internal
         /// storage for the vertices.  Thus modifying the array
         /// may not modify the geometry itself.
-        /// Use the <see cref="CoordinateSequence.SetOrdinate"/> method
+        /// Use the <see cref="CoordinateSequence.SetOrdinate(int, int, double)"/> or
+        /// <see cref="CoordinateSequence.SetOrdinate(int, Ordinate, double)"/> method
         /// (possibly on the components) to modify the underlying data.
         /// If the coordinates are modified,
         /// <see cref="Geometry.GeometryChanged"/> must be called afterwards.
@@ -362,7 +363,8 @@ namespace NetTopologySuite.Geometries
         /// </remarks>
         /// <returns>The vertices of this <c>Geometry</c>.</returns>
         /// <seealso cref="Geometry.GeometryChanged"/>
-        /// <seealso cref="CoordinateSequence.SetOrdinate"/>
+        /// <seealso cref="CoordinateSequence.SetOrdinate(int, int, double)"/>
+        /// <seealso cref="CoordinateSequence.SetOrdinate(int, Ordinate, double)"/>
         public abstract Coordinate[] Coordinates { get; }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/NetTopologySuite/Geometries/Geometry.cs
```diff
@@ -1515,24 +1517,7 @@ public virtual Geometry Reverse()
...
-                var g2 = other;
-                return GeometryCollectionMapper.Map(
-                    (GeometryCollection)this, g => g.Intersection(g2));
-            }
-
-            // No longer needed since GCs are handled by previous code
-            //CheckNotGeometryCollection(this);
-            //CheckNotGeometryCollection(other);
-            return SnapIfNeededOverlayOp.Overlay(this, other, SpatialFunction.Intersection);
+            return Factory.GeometryOverlay.Intersection(this, other);
```
airbreather, line 1520: I'd like it if this (and the other similar instance methods) were to check and throw an exception if `!Equals(Factory.GeometryOverlay, other.Factory.GeometryOverlay)`.

This isn't a problem in JTS, since there's only one overlay implementation that's allowed to exist in the entire package at any given time, but for us, we **do** allow there to be more than one, so it could happen.

It's my opinion that `a.Intersection(b)`, `a.Union(b)`, and `a.SymmetricDifference(b)` should always behave practically identically to `b.Intersection(a)`, `b.Union(a)`, and `b.SymmetricDifference(a)` (plus-or-minus minor differences related to numerical robustness).  As-written, there could be substantial differences between the two, because they could use significantly different algorithms, which could perhaps surface as a "gotcha" in some use cases that we don't know about at the moment, which could add another layer of difficulty when debugging.

As for `Difference`, I don't have a specific reason to want it to throw an exception if the `GeometryOverlay` implementations are different, other than consistency with the other bivariate overlay operations.

src/NetTopologySuite/Geometries/GeometryFactory.cs
```diff
@@ -68,6 +68,26 @@ public class GeometryFactory
...
+            set
+            {
+                if (_geometryOverlay != null)
+                    throw new InvalidOperationException("GeometryOverlay already set.");
+
+                if (value == null)
+                    value = NtsGeometryServices.Instance.GeometryOverlay;
+
+                _geometryOverlay = value;
+            }
```
airbreather, line 88: Now that this is internal, I think we can simplify without losing anything important:
```csharp
get { return _geometryOverlay ?? NtsGeometryServices.Instance.GeometryOverlay; }
set { _geometryOverlay = value; }
```

test/NetTopologySuite.Tests.NUnit/Geometries/GeometryOverlayTest.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            try
+            {
+                TryIntersection();
+                Assert.Fail("Intersection operation should have failed but did not");
+            }
+            catch (TopologyException ex)
+            {
+                // ignore - expected result
+            }
```
airbreather, line 90: I think this is equivalent to:
```csharp
Assert.That(TryIntersection, Throws.InstanceOf<TopologyException>());
```

test/NetTopologySuite.Tests.NUnit/Geometries/GeometryOverlayTest.cs
```diff
@@ -0,0 +1,115 @@
...
+        private void CheckIntersectionSucceeds()
+        {
+            try
+            {
+                TryIntersection();
+            }
+            catch (TopologyException ex)
+            {
+                Assert.Fail("Intersection operation failed.");
+            }
```
airbreather, line 102: I think this is equivalent to:
```csharp
Assert.That(TryIntersection, Throws.Nothing);
```

test/NetTopologySuite.Tests.NUnit/Geometries/GeometryOverlayTest.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        private void TryIntersection()
+        {
+            var a = Read(NtsGeometryServices.Instance.CreateGeometryFactory(),
```
airbreather, line 107: suggestion: add a comment here to find a different failure case if something changes that breaks our "this should fail" test.

test/NetTopologySuite.Tests.NUnit/Geometries/GeometryOverlayTest.cs
```diff
@@ -0,0 +1,115 @@
...
+        public void SetUp()
+        {
+            _ngs = NtsGeometryServices.Instance;
+        }
+
+        [OneTimeTearDown]
+        public void TearDown()
+        {
+            NtsGeometryServices.Instance = _ngs;
+        }
```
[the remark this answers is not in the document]
FObermaier: Changed

test/NetTopologySuite.Tests.NUnit/Geometries/GeometryOverlayTest.cs
FObermaier: dito

src/NetTopologySuite/Geometries/GeometryOverlay.cs
```diff
@@ -0,0 +1,222 @@
...
+    /// <summary>
+    /// A class that encapsulates geometry overlay functionality
+    /// </summary>
+    [Serializable]
+    public abstract class GeometryOverlay
+    {
+        /// <summary>
+        /// Gets a value indicating a geometry overlay operation class that uses old NTS overlay operation set.
+        /// </summary>
+        public static GeometryOverlay Old
```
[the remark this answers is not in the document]
FObermaier: Renamed

src/NetTopologySuite/Geometries/Geometry.cs
FObermaier: Equality check in `GeometryOverlay`

FObermaier:
@airbreather , JTS is about to release 1.18 so it'd be great if you could find some time to re-review this and we can follow suit asap.

src/NetTopologySuite/NtsGeometryServices.cs
```diff
@@ -69,19 +74,8 @@ public static NtsGeometryServices Instance
...
-                if (_geometryOverlay != null)
-                    throw new InvalidOperationException();
-                if (value == null)
-                    value = DefaultGeometryOverlay;
-
-                _geometryOverlay = value;
-            }
-        }
+        /// <returns>A set of geometry overlay functions.</returns>
+        internal GeometryOverlay GeometryOverlay { get; }
```
airbreather, line 78: IMO this was fine as `public` in this class, since this class is pretty much just a bag of properties anyway.

src/NetTopologySuite/NtsGeometryServices.cs
```diff
@@ -69,19 +74,8 @@ public static NtsGeometryServices Instance
         /// <summary>
         /// Gets or sets a value indicating the operations to use for geometry overlay.
```
airbreather, line 75: (nit): `gets or sets` should just be `gets` now

src/NetTopologySuite/IO/WKTReader.cs
```diff
@@ -61,9 +74,8 @@ public class WKTReader
         /// <param name="geometryFactory">The factory used to create <c>Geometry</c>s.</param>
         public WKTReader(GeometryFactory geometryFactory)
         {
-            _coordinateSequencefactory = geometryFactory.CoordinateSequenceFactory;
-            _precisionModel = geometryFactory.PrecisionModel;
-            DefaultSRID = geometryFactory.SRID;
+            _ntsGeometryServices = NtsGeometryServices.Instance;
+            Factory = geometryFactory;
```
airbreather, line 78: this can cause needless confusion and rework when someone passes in a `GeometryFactory` that they created with a custom `NtsGeometryServices` instance that they use consistently without ever bothering to set it as `NtsGeometryServices.Instance`.

it's also confusing to have both constructors, because we don't give any real direction to show how to use it... **minimally**, I think, we should mark this `GeometryFactory` constructor obsolete if we're going to have both, but we can do better to make the upgrade smoother for people coming from earlier versions.

suggestion (most consistent with how this class has always operated):
- remove the new `NtsGeometryServices` constructor
- add another overload to `NtsGeometryServices.CreateGeometryFactory` that lets us pass in a `GeometryOverlay` instance (and all the baggage that goes with it)
- instead of adding the `_ntsGeometryServices` field and the new constructor, let's add a `_geometryOverlay` field like the others and extract it from the `GeometryFactory` / pass it through in the `Factory` property getter to `NtsGeometryServices.Instance.CreateGeometryFactory` like we currently pass through the others
- `Factory` property setter can now extract the `GeometryOverlay` into the field instead of throwing if it's different.

another way to do this (most consistent with how I would develop it in the first place):
- stop extracting all the individual pieces from the `GeometryFactory` or `NtsGeometryServices` instance that we were passed in (`CoordinateSequenceFactory`, `PrecisionModel`, `SRID`).
- instead, fetch all of those and the `GeometryOverlay` instance, when needed, from the `_ntsGeometryServices` field
- in the `GeometryFactory` constructor and `Factory` property setter, create a new `NtsGeometryServices` instance for the `_ntsGeometryServices` field by extracting **all** of the pieces.
    - this would require `_ntsGeometryServices` to no longer be `readonly`.
    - `Factory` property setter can lose the `GeometryOverlay` check now.
- mark the `GeometryFactory` constructor (and the `Factory` property, I guess) `[Obsolete]`, directing callers to use the `NtsGeometryServices` constructor instead (we did this part with `WKBReader` too, it seems)

combining what I feel are the good parts of both (this is what I would personally aim to do here):
- stop extracting all the individual pieces from the `GeometryFactory` or `NtsGeometryServices` instance that we were passed in (`CoordinateSequenceFactory`, `PrecisionModel`, `SRID`) 
- instead, fetch all of those and the `GeometryOverlay` instance when needed, from the `_ntsGeometryServices` field
- in the `GeometryFactory` constructor and `Factory` property setter, create a new `NtsGeometryServices` instance for the `_ntsGeometryServices` field by extracting **all** of the pieces.  this would require `_ntsGeometryServices` to no longer be `readonly`.
- mark the `GeometryFactory` constructor and the `Factory` property `[Obsolete]`, directing callers to use the `NtsGeometryServices` constructor instead (we did this part with `WKBReader` too, it seems)
- add another overload to `NtsGeometryServices.CreateGeometryFactory` that lets us pass in a `GeometryOverlay` instance (and all the baggage that goes with it)
- in the `GeometryFactory` getter, change `NtsGeometryServices.Instance.CreateGeometryFactory(stuff)` to `_ntsGeometryServices.CreateGeometryFactory()`

airbreather (changes requested):

src/NetTopologySuite/IO/WKTReader.cs
airbreather: To clarify: one example reason why someone might be setting `NtsGeometryServices.Instance` differently from what they're using elsewhere would be if they're an ASP.NET Core Web API application that gives its downstream users the same choice that we give our own downstream users: "do you want to opt-in to trying the new stuff for this request, knowing that it might change outputs from earlier versions of this application?"

FObermaier: I like the combination.

src/NetTopologySuite/IO/WKTReader.cs
```diff
@@ -104,17 +98,16 @@ public bool IsOldNtsMultiPointSyntaxAllowed
...
-                        throw new ArgumentException("Factory has incompatible set of overlay functions assigned.", nameof(value));
-
-                    _coordinateSequencefactory = value.CoordinateSequenceFactory;
-                    _precisionModel = value.PrecisionModel;
-                    DefaultSRID = value.SRID;
+                    _ntsGeometryServices = (NtsGeometryServices)Activator.CreateInstance(_ntsGeometryServices.GetType(),
+                        value.CoordinateSequenceFactory, value.PrecisionModel,
+                        value.SRID, value.GeometryOverlay);
+                    // Not sure about this:
+                    // DefaultSRID = value.SRID;
```
airbreather, line 110: why is this different from what we do in the `GeometryFactory` constructor?

src/NetTopologySuite/IO/WKTReader.cs
```diff
@@ -104,17 +98,16 @@ public bool IsOldNtsMultiPointSyntaxAllowed
...
             {
                 if (value != null)
                 {
-                    if (value.GeometryOverlay != _ntsGeometryServices.GeometryOverlay)
-                        throw new ArgumentException("Factory has incompatible set of overlay functions assigned.", nameof(value));
-
-                    _coordinateSequencefactory = value.CoordinateSequenceFactory;
-                    _precisionModel = value.PrecisionModel;
-                    DefaultSRID = value.SRID;
+                    _ntsGeometryServices = (NtsGeometryServices)Activator.CreateInstance(_ntsGeometryServices.GetType(),
```
airbreather, line 106: reflection is dangerous; we can't force external subclasses to have a constructor that looks like this, we can't advertise that requirement strongly enough (people won't read the documentation, and only this very unlikely use case would care, so it would go unnoticed until a random time in the future), and it's not really **that** critical to our operation, right?

why not do the same that the `GeometryFactory` constructor does, and mark this property `[Obsolete]`?

src/NetTopologySuite/IO/WKTReader.cs
FObermaier: Because otherwise `GeometryFactoryExTest` would fail.
It is not possible to configure `NtsGeometryServices` to return anything other than `GeometryFactory`.

We could catch the Exception and provide necessary information.

airbreather:
Branch https://github.com/NetTopologySuite/NetTopologySuite/tree/feature/airbreather-fixes has my proposed fixes:
1. Remove the new use of reflection in `WKTReader`
1. Mark properties obsolete if they involve anything other than using the `NtsGeometryServices` instance that was passed-in to the constructor
1. Mark `DefaultSRID` obsolete (caller can provide a `NtsGeometryServices` instance if they have a different opinion of what the SRID should be)
1. Ensure tests don't fail after doing the above
1. Add some null checks, after the last changes started triggering a test failure

All NUnit tests pass

FObermaier:
@airbreather, for simplicity I merged your branch. Thanks for your input!

PR #483, opened 2021-02-09
Add CoordinateEqualityComparer 

FObermaier:
This PR adds `CoordiateEqualityComparer` class. It is added as a property to `NtsGeometryServices`.
Additional changes:
* Add creating `NtsGeometryServices` as readonly field to `GeometryFactory` (`_services`).
* Add pass through property `CoordinateEqualityComparer` via `_services.CoordinateEqualityComparer`
* Remove `GeometryOverlay` instance field from `GeometryFactory`
* Add pass through property `GeometryOverlay` via `_services.GeometryOverlay`
* Remove `SerializableAttribute` from GeometryOverlay

Still ToDo:
- [x] Remove `SerializableAttribute` from CoordinateEqualityComparer
- [x] Remove ?? coalesce when accessing `_services`

relates to #447


FObermaier:
@airbreather, I'd like to see this resolved so we can get `GeometryOverlay` out.
Thanks

src/NetTopologySuite/Geometries/CoordinateEqualityComparer.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// An implementation of <see cref="CoordinateEqualityComparer"/> that includes
+    /// <see cref="Coordinate.Z"/> into equality tests.
+    /// </summary>
+    [Serializable]
+    public sealed class CoordinateZComparer : CoordinateEqualityComparer
+    {
+        /// <inheritdoc cref="CoordinateEqualityComparer.AreEqual"/>
+        public override bool AreEqual(Coordinate a, Coordinate b, double tolerance)
+        {
+            if (tolerance.Equals(0d))
```
airbreather, line 52: might as well use `== 0` here like elsewhere

src/NetTopologySuite/Geometries/CoordinateEqualityComparer.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// An implementation of <see cref="CoordinateEqualityComparer"/> that includes
+    /// <see cref="Coordinate.Z"/> and <see cref="Coordinate.M"/> into equality tests.
+    /// </summary>
+    [Serializable]
+    public sealed class CoordinateZMComparer : CoordinateEqualityComparer
+    {
+        /// <inheritdoc cref="CoordinateEqualityComparer.AreEqual"/>
+        public override bool AreEqual(Coordinate a, Coordinate b, double tolerance)
+        {
+            if (tolerance.Equals(0d))
```
airbreather, line 79: might as well use `== 0` here like elsewhere

src/NetTopologySuite/Geometries/GeometryFactory.cs
```diff
@@ -99,21 +110,22 @@ public static Point CreatePointFromInternalCoord(Coordinate coord, Geometry exem
         /// <param name="precisionModel">A precision model</param>
         /// <param name="srid">A spatial reference id</param>
         /// <param name="coordinateSequenceFactory">A coordinate sequence factory</param>
-        /// <param name="geometryOverlay">A geometry overlay function set</param>
-        public GeometryFactory(PrecisionModel precisionModel, int srid, CoordinateSequenceFactory coordinateSequenceFactory, GeometryOverlay geometryOverlay)
+        /// <param name="services">A geometry overlay function set</param>
```
airbreather, line 113: doc is outdated now

src/NetTopologySuite/Geometries/CoordinateEqualityComparer.cs
```diff
@@ -0,0 +1,98 @@
...
+    [Serializable]
+    public sealed class CoordinateZComparer : CoordinateEqualityComparer
+    {
+        /// <inheritdoc cref="CoordinateEqualityComparer.AreEqual"/>
+        public override bool AreEqual(Coordinate a, Coordinate b, double tolerance)
+        {
+            if (tolerance.Equals(0d))
+                return a.Equals(b) && a.Z.Equals(b.Z);
+
+            return a.Distance(b) <= tolerance && DistanceInZ(a, b) <= tolerance;
```
airbreather, line 55: :page_facing_up: it's a bit weird to see the same tolerance used for both xy distance and Z... is there any reason for using that same tolerance?  it seems, perhaps, safest to just do:
```csharp
return base.AreEqual(a, b, tolerance) && a.Z.Equals(b.Z);
```

src/NetTopologySuite/Geometries/CoordinateEqualityComparer.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        /// <inheritdoc cref="CoordinateEqualityComparer.AreEqual"/>
+        public override bool AreEqual(Coordinate a, Coordinate b, double tolerance)
+        {
+            if (tolerance.Equals(0d))
+                return a.Equals(b) && a.Z.Equals(b.Z) && a.M.Equals(b.M);
+
+            return a.Distance(b) <= tolerance &&
+                   Distance(a.Z, b.Z) <= tolerance &&
+                   Distance(a.M, b.M) <= tolerance;
```
airbreather, line 84: :page_facing_up: as above, it's a bit weird to see the same tolerance used for xy distance, Z, and M... is there any reason for using that same tolerance?  it seems, perhaps, safest to just do:
```csharp
return base.AreEqual(a, b, tolerance) && a.Z.Equals(b.Z) && a.M.Equals(b.M);
```

src/NetTopologySuite/NtsGeometryServices.cs
```diff
@@ -104,13 +105,29 @@ public NtsGeometryServices(CoordinateSequenceFactory coordinateSequenceFactory,
...
         /// <param name="geometryOverlay">The geometry overlay function set to use.</param>
-        public NtsGeometryServices(CoordinateSequenceFactory coordinateSequenceFactory, PrecisionModel precisionModel, int srid, GeometryOverlay geometryOverlay)
+        public NtsGeometryServices(CoordinateSequenceFactory coordinateSequenceFactory, PrecisionModel precisionModel, int srid, GeometryOverlay geometryOverlay):
+            this(coordinateSequenceFactory, precisionModel, srid, geometryOverlay, new CoordinateEqualityComparer())
+        { }
+
+        /// <summary>
+        /// Creates an instance of this class, using the provided <see cref="CoordinateSequenceFactory"/>,
+        /// <see cref="PrecisionModel"/>, a spatial reference Id (<paramref name="srid"/>) and
+        /// a <see cref="Geometries.GeometryOverlay"/>.
```
airbreather, line 115: `and a <see cref="Geometries.CoordinateEqualityComparer"/>`

src/NetTopologySuite/NtsGeometryServices.cs
```diff
@@ -131,6 +148,13 @@ public static NtsGeometryServices Instance
         /// <returns>A set of geometry overlay functions.</returns>
         public GeometryOverlay GeometryOverlay { get; }
 
+        /// <summary>
+        /// Gets or sets an (Optional) object that is used to test 2 coordinates for equality.
+        /// <br/> If <c>null</c>, standard JTS-like equality checks are performed.
+        /// </summary>
+        /// <returns>A coordinate equality tester object or <c>null</c></returns>
```
airbreather, line 155: these suggest it can be `null`, but it actually can't be, right?

test/NetTopologySuite.Samples.Console/Tests/Github/Issue447.cs
```diff
@@ -0,0 +1,33 @@
...
+            var pt2 = i.CreateGeometryFactory().CreatePoint(new CoordinateZ(1, 2, 3));
+            Assert.IsFalse(pt1 == pt2);
+            Assert.AreNotEqual(pt1, pt2);
+        }
+
+        [Test]
+        public void Test2()
+        {
+            var pt1 = new NetTopologySuite.Geometries.Point(new NetTopologySuite.Geometries.Coordinate(10, 10));
+            var pt2 = new NetTopologySuite.Geometries.Point(new NetTopologySuite.Geometries.CoordinateZ(10, 10));
```
airbreather, line 27: we should still use a `GeometryFactory.CreatePoint` method instead of the constructors, though?

if for no other reason than to be explicit

src/NetTopologySuite/NtsGeometryServices.cs
```diff
@@ -104,13 +105,29 @@ public NtsGeometryServices(CoordinateSequenceFactory coordinateSequenceFactory,
         /// <param name="precisionModel">The precision model.</param>
         /// <param name="srid">The default spatial reference ID</param>
         /// <param name="geometryOverlay">The geometry overlay function set to use.</param>
-        public NtsGeometryServices(CoordinateSequenceFactory coordinateSequenceFactory, PrecisionModel precisionModel, int srid, GeometryOverlay geometryOverlay)
+        public NtsGeometryServices(CoordinateSequenceFactory coordinateSequenceFactory, PrecisionModel precisionModel, int srid, GeometryOverlay geometryOverlay):
+            this(coordinateSequenceFactory, precisionModel, srid, geometryOverlay, new CoordinateEqualityComparer())
+        { }
```
airbreather, line 110: since we haven't released with `GeometryOverlay` yet, I think we can get away with removing this constructor and just having the only new constructor in this release accept both parameters?

src/NetTopologySuite/Geometries/CoordinateEqualityComparer.cs
FObermaier: Or make base implementation check tolerance for each ordinate instead of distance, and
Create a derived NTS implementation that checks like done before.

test/NetTopologySuite.Samples.Console/Tests/Github/Issue447.cs
```diff
@@ -13,18 +14,78 @@ public class Issue447
...
+        }
+
+        private static double Increment(double value)
+        {
+            return Math.Sign(value) * BitConverter.Int64BitsToDouble(BitConverter.DoubleToInt64Bits(Math.Abs(value)) + 1);
+        }
+        private static double Decrement(double value)
+        {
+            return Math.Sign(value) * BitConverter.Int64BitsToDouble(BitConverter.DoubleToInt64Bits(Math.Abs(value)) - 1);
         }
```
airbreather: I think multiplying by `Math.Sign` might be wrong for negative values (which we don't have, so we wouldn't know)... regardless, we can use [Math.BitIncrement](https://docs.microsoft.com/en-us/dotnet/api/system.math.bitincrement) and [Math.BitDecrement](https://docs.microsoft.com/en-us/dotnet/api/system.math.bitdecrement) in our tests these days.

airbreather (approved):

test/NetTopologySuite.Samples.Console/Tests/Github/Issue447.cs
FObermaier: 👍 . I'm always amazed how well you know the dotnet api and its libraries.
Thanks

airbreather:
FWIW, `NetTopologySuite.Samples.Console/Tests/Github/Issue474Fixture.cs` also needs the comparer on L42

PR #485, opened 2021-02-10
Deviate from JTS in extra-dimension coordinate handling.

### LINKED ISSUE #461: Proposed deviation from JTS on resolution to locationtech/jts#580
@FObermaier

(cc: @dr-jts in case you have any thoughts about this topic)

Given that the only actual **problem** stated in locationtech/jts#580 is that users would lose some of their coordinate data when creating `Coordinate` instances, I propose reverting b713c2b and fixing that issue by removing `PackedCoordinateSequence`'s override of `GetCoordinateCopy` instead.

This is because the default behavior in NTS, when not overridden, will return an instance that contains all ordinate values, even if the user accidentally asked for 4 spatial dimensions.  We have specific tests for this behavior, so we might as well not override it just for the sake of making it only support a narrower set of use cases.  Specific tests are:
https://github.com/NetTopologySuite/NetTopologySuite/blob/17684339365c57f429b173b3d82b2a1ff9d84b78/test/NetTopologySuite.Tests.NUnit/Geometries/Implementation/BaseCoordinateSequenceTest.cs#L255-L273

Given that this would fix the only actual reported **problem** from that issue, I don't think it's right to change the behavior of `CoordinateSeqeuenceFactory.Create(int size, int dimension)`, since this is a break from 2.0 / 2.1...

airbreather:
- 4th spatial dimension is now accessible in all cases.
- Certain coordinate sequences now deviate from JTS to support more cases.

Resolves #461

### DIFF
--- a/src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequenceFactory.cs
+++ b/src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequenceFactory.cs
@@ -36,6 +36,8 @@ namespace NetTopologySuite.Geometries.Implementation
         {
             int spatial = dimension - measures;
 
+            // DEVIATION: JTS can't create Coordinate instances other than XY, XYZ, XYM, or XYZM.
+            // we can, so no need to clip spatial / measures.
             if (spatial < 2)
             {
                 spatial = 2; // handle bogus dimension
--- a/src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequenceFactory.cs
+++ b/src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequenceFactory.cs
@@ -36,17 +36,6 @@ namespace NetTopologySuite.Geometries.Implementation
         {
             int spatial = dimension - measures;
 
-            if (measures > 1)
-            {
-                measures = 1; // clip measures
-            }
-
-            if (spatial > 3)
-            {
-                spatial = 3; // clip spatial dimension
-                // throw new ArgumentException("spatial dimension must be <= 3");
-            }
-
             if (spatial < 2)
             {
                 spatial = 2; // handle bogus dimension
--- a/src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
+++ b/src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
@@ -111,6 +111,10 @@ namespace NetTopologySuite.Geometries.Implementation
         [Obsolete("Unused.  This will be removed in a future release.")]
         protected virtual Coordinate GetCoordinateInternal(int index)
         {
+            // DEVIATION: JTS can't create Coordinate instances other than XY, XYZ, XYM, or XYZM.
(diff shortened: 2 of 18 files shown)


src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequenceFactory.cs
```diff
@@ -36,17 +36,6 @@ public override CoordinateSequence Create(int size, int dimension, int measures)
         {
             int spatial = dimension - measures;
 
-            if (measures > 1)
```
FObermaier, line 39: We should note that we deliberatly deviate from JTS' CoordinateArraySequence.

src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
```diff
@@ -43,43 +43,7 @@ public sealed override Coordinate GetCoordinate(int i)
             var arr = GetCachedCoords();
             if(arr != null)
                  return arr[i];
-            return GetCoordinateInternal(i);
```
FObermaier, line 46: We should note that we deliberatly deviate from JTS' PackedCoordinateSequence.

FObermaier: Perhaps put it in `#region ... end region` and comment the code.

src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
```diff
@@ -132,69 +96,6 @@ private Coordinate[] GetCachedCoords()
             return null;
         }
 
-        /// <summary>
```
FObermaier, line 135: dito

FObermaier (approved):
Consider my annotations as suggestion.
I think it is easier to maintain this code if we are remembered that for some reason or other we deliberatly deviate from the JTS' original.

FObermaier:
@airbreather, another idea is to place a README.md file into `NetTopologySuite.Geometries.Implementation` stating that the content of the whole namespace is different than `org.locationtech.jts.geom.impl` and that changes over there should only be ported if the related unit test fails on NTS, too.

I'd like to release v2.2 soon, that is why i assigned some issues to milestone 2.2. I'd be grateful if you could have a look at those and tell me if you'd like to see any of these addressed before a new release. Thanks!

airbreather:
Ugh.  I was trying to get the initialization from `Coordinate[]` to consistently take the highest dimension and measures from the given array, but it actually seems like using `CoordinateArrays.Dimension` and `CoordinateArrays.Measures` separately is the wrong thing to do.

What we really need is something that tells us the smallest (dimension, measures) that can hold all the same slots from the given coordinate array.

As-is, a `Coordinate[]` with some `CoordinateZ` and some `CoordinateM` would tell us to use dimension 3 and measures 1, but we really would need dimension 4 and measures 1... I'll roll some of that last commit back and have everything keep looking at just the first coordinate.

FObermaier (approved):
We should not use the obsolete constructors ourselves. Other than that I'm fine with this.

PR #500, opened 2021-02-13
New feature: Calculates a point along the curve at a given distance.

nakulLukan:
### Description
Calculates a point (coordinate) along a curve (LineString) from the start point at a given distance (in meters).

### DIFF
--- a/src/NetTopologySuite/Geometries/LineString.cs
+++ b/src/NetTopologySuite/Geometries/LineString.cs
@@ -1,7 +1,7 @@
-using System;
-using System.Collections.Generic;
 using NetTopologySuite.Operation;
 using NetTopologySuite.Utilities;
+using System;
+using System.Collections.Generic;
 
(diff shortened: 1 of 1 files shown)


airbreather:
:question: doesn't the `NetTopologySuite.LinearReferencing` namespace already have something for this?

FObermaier:
Yes:
https://github.com/NetTopologySuite/NetTopologySuite/blob/f0527264d37b9b0c7a2b5c8b1899d1c2100620ac/src/NetTopologySuite/LinearReferencing/LengthIndexedLine.cs#L47-L64

Boils down to
```C#
var lil = new NetTopologySuite.LinearReferencing.LengthIndexedLine(<linestring from somewhere>);
var pt = lil.ExtractPoint(<position along linestring>, <distance>);
```

nakulLukan:
Thanks @airbreather and @FObermaier. I failed to find the this function so thought of creating one.. :)

PR #502, opened 2021-02-17
Setting SRID copies the factory more faithfully.

### LINKED ISSUE #437: GmlReader creates a GeometryFactory even when initialized with a GeometryFactoryEx
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of NetTopologySuite
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/NetTopologySuite/NetTopologySuite/issues) and [closed](https://github.com/NetTopologySuite/NetTopologySuite/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When using GmlReader to read a geometry and the geometry has an srsName set, it tries to create a GeometryFactory with the correct srid, which is fine. But when I have constructed it with a GeometryFactoryEx with `OrientationOfExteriorRing = LinearRingOrientation.CounterClockwise` then I expect the polygon to be created with a counter clockwise exterior ring, but it is not.

### Steps to Reproduce
```var factory = new GeometryFactoryEx() { OrientationOfExteriorRing = LinearRingOrientation.CounterClockwise };
var reader = new GmlReader(factory);
var geom = (Polygon)reader.Read(_xmlReader);
```

my gml looks like this:
`<gml:Polygon srsName="urn:ogc:def:crs:EPSG::25832"><gml:exterior><gml:LinearRing><gml:posList>723814.5274999999 6174591.989800001 [...]`

### System Configuration
Tell us about the environment where you are experiencing the bug:
- NetTopologySuite version: 2.1.0
- Other NetTopologySuite packages and versions: NetTopologySuite.IO.SqlServerBytes 2.0.0
- Environment (Operating system, version and so on): Windows 10
- .NET Framework version: dotnet core 3.1
- Additional information:

_Thanks for reporting the issue to NetTopologySuite!_

airbreather:
GeometryFactory now has a `WithSRID` method, which `GeometryFactoryEx` overrides.
Base class still uses `NtsGeometryServices` to share instances where possible.
`GmlReader` uses this method too, so this resolves #437

### DIFF
--- a/src/NetTopologySuite/Geometries/GeometryFactoryEx.cs
+++ b/src/NetTopologySuite/Geometries/GeometryFactoryEx.cs
@@ -108,7 +108,8 @@ namespace NetTopologySuite.Geometries
         /// </summary>
         /// <remarks>
         /// The setter of this property has to be used prior to any call
-        /// to <c>CreatePolygon</c> or <c>"CreateMultiPolygon</c></remarks>
+        /// to <c>CreatePolygon</c>, <c>CreateMultiPolygon</c>, or
+        /// <c>ReplaceSRID</c></remarks>
         /// <seealso cref="GeometryFactory.CreatePolygon(Coordinate[])"/>
         /// <seealso cref="GeometryFactory.CreatePolygon(CoordinateSequence)"/>
         /// <seealso cref="GeometryFactory.CreatePolygon(LinearRing)"/>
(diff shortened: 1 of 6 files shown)


src/NetTopologySuite/Geometries/GeometryFactoryEx.cs
```diff
@@ -205,6 +206,21 @@ public override MultiPolygon CreateMultiPolygon(Polygon[] polygons)
...
         }
 
+        /// <inheritdoc />
+        public override GeometryFactory WithSRID(int srid)
+        {
+            var clone = base.WithSRID(srid);
+            if (!(clone is GeometryFactoryEx cloneEx))
+            {
+                cloneEx = new GeometryFactoryEx(PrecisionModel, srid, CoordinateSequenceFactory, GeometryServices);
+            }
```
airbreather, line 216: actually, let's always do the explicit copy... base version via `NtsGeometryServices` might return a shared instance that we shouldn't be messing with.

FObermaier (approved):

PR #514, opened 2021-04-11
Add a lower-level CoordinateSequence implementation

### LINKED ISSUE #482: Customizable Variant of PackedDoubleCoordinateSequence
Discussing with @bjornharrtell on Gitter, it seems like there's a need for a generalized variant of `DotSpatialAffineCoordinateSequence` that supports (minimally) `ArraySegment<double>` rather than forcing the data to exactly cover an entire array.

In my own research, I've also found a use for a `CoordinateSequence` implementation that supports a memory layout where X and Y are in their own parallel arrays (particularly for pointwise SIMD operations).

Even `PackedDoubleCoordinateSequence`-style layouts would benefit from being able to be created from slices of existing arrays instead of always having to fully populate the array.

To that end, I'm going to get started on a coordinate sequence (+ factory) that supports more-or-less any densely packed layout backed by `Memory<double>` instances.  The requirements will be, roughly:
1. For any given dimension, the entire contents must fit within a single `Memory<double>`.
1. No gaps or scrambling: `array[0]` must hold the first value in one dimension, `array[1]` must hold either the first value in another dimension or (if the array holds data for only a single dimension) the second value in the first dimension, and so on.

airbreather:
Note that "lower-level" does not necessarily mean that typical usage will be faster than when using other sequence types.  In fact, many use cases will see slower accesses, on average.

Rather, this sequence type is intended to provide callers with more flexibility to manage their own memory and data layout, reducing the number of situations where data must be copied in order to produce and consume the sequence.

The only known "blind spot" where copying is still needed even in sequential memory layouts is in situations where the underlying coordinate data is mixed in with non-coordinate data, e.g., if the sequence looks something like:
```
[x0, y0, z0, OTHER_STUFF, x1, y1, z1, OTHER_STUFF, ..., xn, yn, zn, OTHER_STUFF]
```

Otherwise, as long as your coordinate data is all laid out consistently and sequentially in virtual memory, this sequence should work for you.  You **may** need to implement your own `MemoryManager<double>` to make it work for you (e.g., if you produce and consume the entire sequence data in stack-only memory), but if you have something like `Memory<byte>` that you just need to convert to `Memory<double>`, check out [`MemoryExtensions.Cast<TFrom, TTo>`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.highperformance.memoryextensions.cast?view=win-comm-toolkit-dotnet-7.0) from the [`Microsoft.Toolkit.HighPerformance`](https://www.nuget.org/packages/Microsoft.Toolkit.HighPerformance) package.

Resolves #482.

### DIFF
--- a/src/NetTopologySuite/Geometries/Implementation/RawCoordinateSequence.cs
+++ b/src/NetTopologySuite/Geometries/Implementation/RawCoordinateSequence.cs
@@ -1,0 +1,435 @@
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.Serialization;
+
+namespace NetTopologySuite.Geometries.Implementation
+{
+    /// <summary>
+    /// An implementation of <see cref="CoordinateSequence"/> that packs its contents in a way that
+    /// can be customized by the creator.
+    /// </summary>
+    [Serializable]
+    public sealed class RawCoordinateSequence : CoordinateSequence, ISerializable
+    {
+        private readonly (Memory<double> Array, int DimensionCount)[] _rawData;
+
+        private readonly (int RawDataIndex, int DimensionIndex)[] _dimensionMap;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RawCoordinateSequence"/> class.
+        /// </summary>
+        /// <param name="rawData">
+        /// Contains the raw data for this sequence.
+        /// </param>
+        /// <param name="dimensionMap">
+        /// Contains a pair of indexes to tell us, for each dimension, where to find its data in
+        /// <paramref name="rawData"/>.
+        /// </param>
+        /// <param name="measures">
+        /// The value for <see cref="CoordinateSequence.Measures"/>.
+        /// </param>
+        public RawCoordinateSequence(Memory<double>[] rawData, (int RawDataIndex, int DimensionIndex)[] dimensionMap, int measures)
+            : base(GetCountIfValid(rawData, dimensionMap), dimensionMap.Length, measures)
+        {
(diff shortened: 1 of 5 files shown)


FObermaier (approved):

TranstecJasonD:
Regarding the blind spot mentioned. PROJ has a nice solution for this. You define a stride, which is how far apart each element is in bytes. This way sequences like above can be handled. See the documentation for proj_trans_generic:

https://proj.org/development/reference/functions.html#coordinate-transformation

airbreather:
> Regarding the blind spot mentioned. PROJ has a nice solution for this. You define a stride, which is how far apart each element is in bytes. This way sequences like above can be handled. See the documentation for proj_trans_generic:
> 
> [proj.org/development/reference/functions.html#coordinate-transformation](https://proj.org/development/reference/functions.html#coordinate-transformation)

Yep -- I'm aware of this.  The main problem that I had with doing it this way is that it would be infectious: everything that interacts with this would have to work in terms of bytes instead of sequences of `double` values and remember to multiply or divide by 8, and they would need to cast from `Memory<byte>` to `Memory<double>`, etc., which is not great for usability.

While this is admittedly an advanced API in general, and therefore it can be more acceptable to sacrifice ease of use in exchange for things like closing this "blind spot", exposing everything as `Memory<double>` and counting elements instead of bytes seems **significantly** more idiomatic in the .NET world than the alternative.

Do you have a use case where x/y data is embedded alongside other data in arrays of arbitrary unmanaged structs?  If so, I wouldn't mind spending a bit of time to write up an experimental version that will work with `Memory<byte>` and arbitrary strides.

TranstecJasonD:
I personally do not have a use case for this scenario, just mentioning a possible solution. I almost, if not always, keep my x-y data in simple arrays.

PR #551, opened 2021-09-30
Improve CoordinateSequences.Copy

FObermaier:
Instead of copying sequences on a per Coordinate basis this implementation attempts to copy underlying arrays as a whole.

### DIFF
--- a/src/NetTopologySuite/Geometries/CoordinateSequences.cs
+++ b/src/NetTopologySuite/Geometries/CoordinateSequences.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.Text;
+using NetTopologySuite.Geometries.Implementation;
 using NetTopologySuite.IO;
 
 namespace NetTopologySuite.Geometries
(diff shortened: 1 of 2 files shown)


src/NetTopologySuite/Geometries/CoordinateSequences.cs
```diff
@@ -54,10 +56,179 @@ public static void Swap(CoordinateSequence seq, int i, int j)
...
+        /// <param name="seq0">A <c>CoordinateSequence</c></param>
+        /// <param name="seq1">A <c>CoordinateSequence</c></param>
+        /// <param name="seq0Indices">The array of common ordinate indices as in <paramref name="seq0"/></param>
+        /// <param name="seq1Indices">The array of common ordinate indices as in <paramref name="seq1"/></param>
+        private static void GetCommonOrdinateIndices(CoordinateSequence seq0, CoordinateSequence seq1,
+            out int[] seq0Indices, out int[] seq1Indices)
+        {
+            var srcIndexList = new List<int>(16);
+            var destIndexList = new List<int>(16);
+            var commonOrdinates = seq0.Ordinates & seq1.Ordinates;
```
airbreather, line 95: Technically, throughout this file, all this checking of `Ordinates` instead of `Dimension` + `Measures` is incomplete: `Ordinates` only describes the first 16 spatial and the first 16 measures, whereas JTS (and therefore NTS) does actually allow you to describe more than that because `Dimension` + `Measures` are the source of truth for the ordinate values that are stored in any given sequence.

Of course, that's completely insane, so pretty much all users with normal requirements ought to be able to ignore this technicality for their own implementations, but last I checked, JTS never validated that `dimension` is between 2 and 4, nor that `dimension - measures` is between 2 and 3, so the Packed sequences happily allow things to get ridiculous.  I made a couple of deviations from JTS in 2.0 to complete the story, e.g.: https://github.com/NetTopologySuite/NetTopologySuite/blob/4166fff890b6ce28937a54540d96c08370fb4fdf/src/NetTopologySuite/Geometries/Coordinates.cs#L43-L44

and also test cases that ensure that the base class continues to handle these edge cases consistently: https://github.com/NetTopologySuite/NetTopologySuite/blob/4166fff890b6ce28937a54540d96c08370fb4fdf/test/NetTopologySuite.Tests.NUnit/Geometries/Implementation/BaseCoordinateSequenceTest.cs#L17-L24

The most correct way to scan two sequences for common ordinates looks something like this (haven't run it through a compiler yet, so please excuse minor errors):
```csharp
public static void Process(CoordinateSequence seq1, CoordinateSequence seq2)
{
    int minCount = Math.Min(seq1.Count, seq2.Count);
    int minSpatial = Math.Min(seq1.Spatial, seq2.Spatial);
    int minMeasures = Math.Min(seq1.Measures, seq2.Measures);

    for (int i = 0; i < minCount; i++)
    {
        // i: the index of a coordinate that both sequences have.
        for (int dim = 0; dim < minSpatial; dim++)
        {
            // dim: a spatial dimension that both sequences have

            // example of something you can do with it legally:
            double val1 = seq1.GetOrdinate(i, dim);
            seq2.SetOrdinate(i, dim, val1);
        }

        for (int measure = 0; measure < minMeasures; measure++)
        {
            // measure: a measure index that both sequences have
            // cannot be used directly, but:
            int dim1 = seq1.Spatial + measure;
            int dim2 = seq2.Spatial + measure;

            // dim1 and dim2 are the dimensions in their respective sequences that
            // match up with the measure index, so this is legal:
            double val1 = seq1.GetOrdinate(i, dim1);
            seq2.SetOrdinate(i, dim2, val1);
        }
    }
}
```

src/NetTopologySuite/Geometries/CoordinateSequences.cs
```diff
@@ -69,14 +240,22 @@ public static void Copy(CoordinateSequence src, int srcPos, CoordinateSequence d
...
-            for (int dim = 0; dim < minDim; dim++)
+            GetCommonOrdinateIndices(src, dest, out int[] srcIndices, out int[] destIndices);
+            CopyCoord(src, srcPos, srcIndices, dest, destPos, destIndices);
+        }
+
+        private static void CopyCoord(CoordinateSequence src, int srcPos, int[] srcIndices,
+                                      CoordinateSequence dest, int destPos, int[] destIndices)
+        {
+            if (srcIndices.Length != destIndices.Length)
+                throw new ArgumentException(nameof(destIndices));
```
airbreather, line 251: (nit) this constructor doesn't set `ParamName`, but rather it sets the `Message`.

src/NetTopologySuite/Geometries/CoordinateSequences.cs
```diff
@@ -54,10 +56,179 @@ public static void Swap(CoordinateSequence seq, int i, int j)
...
 
+        /// <summary>
+        /// Get the common ordinate indices of two <c>CoordinateSequence</c>s.
+        /// </summary>
+        /// <param name="seq0">A <c>CoordinateSequence</c></param>
+        /// <param name="seq1">A <c>CoordinateSequence</c></param>
+        /// <param name="seq0Indices">The array of common ordinate indices as in <paramref name="seq0"/></param>
+        /// <param name="seq1Indices">The array of common ordinate indices as in <paramref name="seq1"/></param>
+        private static void GetCommonOrdinateIndices(CoordinateSequence seq0, CoordinateSequence seq1,
+            out int[] seq0Indices, out int[] seq1Indices)
```
airbreather, line 91: I think we shouldn't need this method.  The revamped `CoordinateSequence` class is designed so that you can scan very efficiently through its coordinates and dimensions, in order, with simple `for` loops like the ones in my bigger comment in this same method.  No need to try to extract separate lists of things that the two have in common.

In other words, every pair of sequences will have the first 2 spatial dimensions in common, and then the next 0..n spatial dimensions in common until one of them runs out of spatial dimensions, and then 0..n measures in common until one of them runs out of measures (the "measured" dimension might not have the same index on both sides).

airbreather (commented):
Some other specific things of note besides that one big comment.

PR #572, opened 2021-11-22
Compatibility with partial trust callers

LTRData:
A few changes to allow .NET Framework callers that do not run with full trust.
* Removed "unsafe" block and changed to managed bit manipulation instead to make all methods callable at lower trust levels.
* Added "SecurityTransparent" attribute to AssemblyInfo.cs to allow partially trusted callers to call into this dll at their own trust level.
* Changed PrecisionModel.GetHashCode slightly to avoid VerificationException at lower trust levels.

Everything should otherwise work as before and all tests pass like before.

### DIFF
--- a/src/NetTopologySuite/AssemblyInfo.cs
+++ b/src/NetTopologySuite/AssemblyInfo.cs
@@ -9,6 +9,4 @@ using System.Security;
 [assembly: ComVisible(false)]
 [assembly: Guid("6B7EB658-792E-4178-B853-8AEB851513A9")]
 
-[assembly: SecurityTransparent]
-
 [assembly: InternalsVisibleTo("NetTopologySuite.Tests.NUnit, PublicKey=" + Consts.PublicKeyToken)]
--- a/src/NetTopologySuite/AssemblyInfo.cs
+++ b/src/NetTopologySuite/AssemblyInfo.cs
@@ -2,10 +2,13 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Security;
 
 [assembly: AssemblyDelaySign(false)]
 [assembly: CLSCompliant(true)]
 [assembly: ComVisible(false)]
 [assembly: Guid("6B7EB658-792E-4178-B853-8AEB851513A9")]
+
+[assembly: SecurityTransparent]
 
 [assembly: InternalsVisibleTo("NetTopologySuite.Tests.NUnit, PublicKey=" + Consts.PublicKeyToken)]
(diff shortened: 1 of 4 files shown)


airbreather:
I'm not at all familiar with CAS, so it took me a while to come up with my own driver code that I could test this on.  I'm still not sure I'm perfectly addressing your concerns, but...

I've pushed a branch https://github.com/NetTopologySuite/NetTopologySuite/tree/nomerge/partial-trust, where commit fe9e945164d65d8bd61fa7986a79f8057e57d2c7 has a driver and a different way to go about this.  The only change that I needed to make on the NTS side was to add this one attribute to `AssemblyInfo.cs`:
```csharp
[assembly: SecurityRules(SecurityRuleSet.Level2, SkipVerificationInFullTrust = true)]
```

When I run ConsoleApp1 from that commit on my Windows VM, it:
- writes "1" when you run it as-is
- throws an error if you remove only the new attribute that was added to NTS's `AssemblyInfo.cs`
- throws an error if you remove only `SecurityPermissionFlag.SkipVerification` from the `SecurityPermission` line

This **SEEMS** to be enough to fulfill what I understand about your requirements, and I would be OK with pushing that change alone.

In my humble opinion, we should not begin limiting NTS's functionality to only what can be done with ["verifiable"](https://docs.microsoft.com/en-us/dotnet/standard/security/key-security-concepts#type-safety-and-security) code, especially at a time when the C# / .NET ecosystem as a whole is rapidly moving:
- **away** from CAS, with the related APIs even having being [marked obsolete](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/code-access-security-apis-obsolete) as of .NET 5.0, and
- **towards** fundamentally "unverifiable" APIs, such as `Span<T>`, which enable performance optimizations that would not otherwise be possible

LTRData:
I agree that CAS is not something that should limit a library like this these days. Practically speaking the only time it is relevant now is ASP.NET applications running in web hosting without full trust. That is something that still happens in some places and since not all ASP.NET applications are easy to migrate to ASP.NET Core, I think it is reasonable to adjust for it where possible, where it does not introduce other issues or less performance or similar.

In any case, I thought that a reasonable compromise could be to not apply an assembly-level attribute, keep the unsafe block in PrecisionUtility but still move away from using Enum.GetHashCode where it was used. The latter is unnecessary and the way it is implemented in .NET Framework requires unverified code access. A better option is to cast to int, which produce a reasonably good result anyway. This effectively means that WKTReader.Read will work at partial trust but code paths that end up in PrecisionUtility.NumberOfDecimals will still break at runtime.

Issue #469, opened 2021-01-12
GMLWriter 2 doesn't follow the gml Specifications


kristofdegrave:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of NetTopologySuite
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/NetTopologySuite/NetTopologySuite/issues) and [closed](https://github.com/NetTopologySuite/NetTopologySuite/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
The Gml 2.1.2 (http://schemas.opengis.net/gml/2.1.2/geometry.xsd) spec states that a GeomertyCollectionType (MultiPointType, MultiLineStringType, MultiPolygonType) have a required attribute called srsName. This includes the srid of the geometry with the format: `EPSG:<srid>`

### Steps to Reproduce
Write a GeomertyCollectionType to the GmlWriter.
Validate the result against the schema.

### System Configuration
Tell us about the environment where you are experiencing the bug:
- NetTopologySuite version: 2.1.0
- Other NetTopologySuite packages and versions: /
- Environment (Operating system, version and so on): Windows 10, VS2019
- .NET Framework version: .Net Framework 4.7.2 and .Net Core 3.1
- Additional information:

_Thanks for reporting the issue to NetTopologySuite!_

PR #575, opened 2021-12-10
add srsName to GML3 (optional) and all GML2

ArneD:
### Prerequisites

### Description
Before I continue with the rest of the tests I would like your input to see if this could be approved.

This PR contains 2 changes to the GML writers.

1. In our use case we would like to have the GML3 writer output the `srsName` attribute.
2. Based on a previous [PR](https://github.com/NetTopologySuite/NetTopologySuite/issues/469) I saw that `Point`, `LinearString`, `LinearRing`, `Polygon` didn't output the `srsName` while it looks required to me: http://schemas.opengis.net/gml/2.1.2/geometry.xsd .
XSD isn't my strong suit, so might be mistaken. It looks like only members of the `MultiPolygon` for example are optional.

Possible improvements:
Add new `ctor` for `GML3Writer` that would take a pattern as input for example `https://www.opengis.net/def/crs/EPSG/0/{0}`.
This would be vulnerable to injection so would have to check if it's a `urn` or `URI`.
Though this could be easily done by creating your own GML writer and inheriting from `GML3Writer` and overriding `GetSrsName`.

### DIFF
--- a/src/NetTopologySuite/IO/GML2/GMLWriter.cs
+++ b/src/NetTopologySuite/IO/GML2/GMLWriter.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Text;
@@ -23,6 +22,8 @@ namespace NetTopologySuite.IO.GML2
         private const int CoordSize = 200;
         private readonly GMLVersion _gmlVersion;
         private readonly string _gmlNs;
+        private readonly string _gmlSrsName = "srsName";
+        private readonly bool _writeSrsNameAttribute;
 
         /// <summary>
         /// Formatter for double values of coordinates
@@ -33,13 +34,15 @@ namespace NetTopologySuite.IO.GML2
         /// Initializes a new instance of the <see cref="GMLWriter"/> class.
         /// </summary>
         public GMLWriter()
-            : this(GMLVersion.Two) // backwards compatibility.
+            : this(GMLVersion.Two, true) // backwards compatibility.
         {
         }
 
-        internal GMLWriter(GMLVersion gmlVersion)
+        internal GMLWriter(GMLVersion gmlVersion, bool writeSrsNameAttribute)
         {
             _gmlVersion = gmlVersion;
+            _writeSrsNameAttribute = writeSrsNameAttribute;
+
             switch (gmlVersion)
             {
                 case GMLVersion.Two:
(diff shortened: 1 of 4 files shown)


test/NetTopologySuite.Tests.NUnit/IO/GML3/GML3WriterTest.cs
```diff
@@ -31,10 +31,13 @@ public class GML3WriterTest
...
 
+        private static int _geometrySRID = 4326;
+
         [Test]
         public void TestGML3Point()
         {
-            var document = ToGML3(CreatePoint());
+            var geometry = CreatePoint();
+            var document = ToGML3(geometry);
             AssertPoint(document);
```
lvermeulen, line 41: don't need this?

src/NetTopologySuite/IO/GML2/GMLWriter.cs
```diff
@@ -23,6 +22,8 @@ public class GMLWriter
         private const int CoordSize = 200;
         private readonly GMLVersion _gmlVersion;
         private readonly string _gmlNs;
+        private readonly string _gmlSrsName = "srsName";
```
FObermaier, line 25: private const string GmlSrsName = "srsName";

src/NetTopologySuite/IO/GML2/GMLWriter.cs
```diff
@@ -499,12 +494,29 @@ protected int SetByteStreamLength(Point point)
             return InitValue + CoordSize;
         }
 
+        protected virtual void WriteAttributeSrsName(int srid, XmlWriter xmlWriter)
+        {
+            if(_writeSrsNameAttribute)
+                xmlWriter.WriteAttributeString(_gmlSrsName, GetSrsName(srid));
```
FObermaier, line 500: xmlWriter.WriteAttributeString(GmlSrsName, GetSrsName(srid));

test/NetTopologySuite.Tests.NUnit/IO/GML3/GML3WriterTest.cs
```diff
@@ -103,7 +106,7 @@ private static XDocument ToGML3(Geometry geom)
             using var ms = new MemoryStream();
```
FObermaier, line 106: Change signature to
```private static XDocument ToGML3(Geometry geom, bool writeSrsName = true)```

Duplicate at least one of the actual tests where `writeSrsName` is set to `false`

FObermaier (changes requested):
Thanks for the addition, please see comments

Issue #330, opened 2019-08-19
Polygon across Dateline and Intersection


sindizzy:
So I have an ellipse (a polygon geometry) that spans the Dateline. If I build the polygon with the coordinates as-is (namely the longitudes are in the range -180 to +180) it then results with the geometry being Invalid and thus when I intersect the polygon with the points I get the error:

`found non-noded intersection between LINESTRING(-155.32030939830221 83.16240616001781, -153.83095758124398 83.120852884772134) and LINESTRING(178.6792290990447 83.116588616289192, -179.84787091746912 83.158205372008268) [ (-155.06664425130143, 83.155328841178274, NaN, NaN) ]`

To alleviate this I take the polygon x values in the range -180 to +180 and shift them to the range 0 to 360. This gives me a valid geometry and the Intersection produces no exceptions. However the intersection only gives me points in the eastern hemisphere and the points in the western hemisphere are ignored.

Is this expected behavior? Is there anything I can do to process this data properly? Is the only solution to shift the points as well? I'm stumped as to how to proceed.

![image](https://user-images.githubusercontent.com/5298460/63289128-f9440680-c283-11e9-8c7c-2e92edbe379c.png)

FObermaier:
You need to use a `MULTIPOLYGON` composed of parts on the eastern and parts of the western hemisphere. One way to achieve that is to 
1. Translate your polygons vertices +180°,
2. Build the `POLYGON`
3. Split your `POLYGON` using a `LINESTRING(180 100, 180 -100)` and build a `MULTIPOLYGON` of the parts ([see this](https://github.com/NetTopologySuite/NetTopologySuite/blob/develop/test/NetTopologySuite.Samples.Console/Operation/Polygonize/SplitPolygonExample.cs)).
4. Translate the vertices of that `MULTIPOLYGON` -180° and use it for your intersection tests.

Issue #124, opened 2016-09-13
polygon buffer retun empty


sendreams:
some self-intersection polygon do buffer action will return empty polygon.

POLYGON ((1.4749999999994841 -5.15, 30 -5.15, 48.9 -2, 108.1997 -2, 130.25148787313435 0.75647348414179227, 130.25148787313435 0.75647348414179161, 130 0.75, 126.3 0.72, 122.9 0.7, 119.2 0.42, 115.2 0.45, 111 0.29, 106.9 0.23, 102.8 0.2, 98.8 0.12, 94.8 0.04, 90.7 -0.08, 86.5 -0.2, 82.4 -0.42, 78.3 -0.57, 74.1 -0.69, 69.9 -0.78, 65.8 -0.87, 61.7 -1.07, 57.7 -1.09, 53.7 -1.229, 49.5 -1.289, 45.3 -1.369, 41.2 -1.719, 37 -2.409, 32.8 -3.219, 28.6 -3.769, 24.5 -4.089, 20.4 -4.429, 16.3 -4.599, 12.1 -4.759, 8 -4.889, 4 -5.049, 1.4749999999994841 -5.15))

do geometry.buffer(0) will return empty polygon.

sendreams:
[TestAttribute]
        public void TestBuffer() {
            WKTReader rdr = new WKTReader();
            IGeometry a = rdr.Read("POLYGON ((1.4749999999994841 -5.15, 30 -5.15, 48.9 -2, 108.1997 -2, 130.25148787313435 0.75647348414179227, 130.25148787313435 0.75647348414179161, 130 0.75, 126.3 0.72, 122.9 0.7, 119.2 0.42, 115.2 0.45, 111 0.29, 106.9 0.23, 102.8 0.2, 98.8 0.12, 94.8 0.04, 90.7 -0.08, 86.5 -0.2, 82.4 -0.42, 78.3 -0.57, 74.1 -0.69, 69.9 -0.78, 65.8 -0.87, 61.7 -1.07, 57.7 -1.09, 53.7 -1.229, 49.5 -1.289, 45.3 -1.369, 41.2 -1.719, 37 -2.409, 32.8 -3.219, 28.6 -3.769, 24.5 -4.089, 20.4 -4.429, 16.3 -4.599, 12.1 -4.759, 8 -4.889, 4 -5.049, 1.4749999999994841 -5.15))");
            a = a.Buffer(0);
        }

[...]

PR #580, opened 2022-01-05
Split polygons at dateline

sindizzy:
### Prerequisites

### Description
Presents a technique for converting coordinates to a polygon. With the coordinates all in one hemisphere it is a
fairly simple process. However, when the geometry crosses the dateline it becomes a more challenging problem.
This routine attempts to do both. That is to say, it converts coordinates to a poly geometry regardless if the
geometry traverses the dateline or not. If the geometry does cross the dateline then it is split up into multiple
parts. This benefits the user in two ways: first, any GIS will be able to display the geometry in a proper way regardless of map projection and two, the GIS will be able to perform geoprocessing functions on the geometry. Note that feature classes do
not support features with multiple parts that share the same edge. Thus, it is left up to the user on how to
commit the geometries to a feature class. Suggestions are to save the multiple geometries as individual features
or to slightly alter the edges of each geometry so as to save all parts as a single multi-part feature. For the
latter option, the algorithm provides an input to trim edges of parts so that they can be used in a single feature.

The algorithm arose from attempting to convert worldwide airspaces (which can be found in any online aviation db) to polygons. Some of the coordinates result in polygons traversing the dateline and the result is ugly. That is because the GIS is not inherently aware that the polygons cross the dateline. This is what a standard conversion will look like:

RegMap_StdProc

The goal is to split those special polygons so that the airspaces are properly represented:

RegMap_SpcProc

The PR includes a sample input with three processing options. If the item crosses the dateline then you will **always** get back multiple geometries no matter what options you use. The difference is what the dev intends to do with the geometries. The sample input will produce the results discussed below. For easy viewing the map projection has been set to Plate Carrée with the central meridian at 180 degrees (the dateline). The PR is also heavy on the debugs and I can remove those once the NTS devs can take a look at the initial proposal.

If the dev intends to add them as a **single part per feature** then this is what you will get with the standard options:

dateline_sp

If the dev intends to add them as **multiple parts per feature** then its a bit more complicated. Through research and observation it was noted that existing datasets representing worldwide polygons already deal with the multi-part issue by trimming one of the polygons ever so slightly such that the two parts do not mathematically share an edge. From my research this trim gap appears to be connected to the feature class tolerance. The trim gap is a variable that the user must provide but should be made as small as possible. For demonstration purposes the trim gap has been made quite large and this is what it would look like with advanced options:

dateline_mp

Abel


src/NetTopologySuite/NetTopologySuite.csproj
```diff
@@ -31,7 +31,6 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(EnableApiCompat)' == 'true' ">
-    <PackageReference Include="Microsoft.DotNet.ApiCompat" Version="6.0.0-beta.21159.11" PrivateAssets="All" />
```
DGuidi, line 34: I see a potential problem here, because `EnableApiCompat == True` but the reference to the package was removed.
Honestly I don't even know how this package is used, but I think it should be restored

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+    /// way regardless of map projection and two, the GIS will be able to perform geoprocessing functions on the geometry. Note that standard feature
+    /// classes do not support individual features with multiple parts that share the same edge. Thus, it is left up to the user on how to commit
+    /// the geometries to a feature class. Suggestions are to save the multiple geometries as individual features or to slightly alter the edges of
+    /// each geometry so as to save all parts as a single multi-part feature. For the latter option, the algorithm provides an input to trim edges
+    /// of parts so that they can be used in a single feature.
+    /// </summary>
+    public class SplitPolygonAtDateline
+    {
+        [STAThread]
+        public static void main()
```
DGuidi, line 30: maybe this code can be moved to a unit test (along with other sample code, possibly?) I know that other "Technique" items uses this parrern (inherited from JTS code) but basically this means that this code isn't tested.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+        {
+            // ========================================================================================
+            // inputs
+            // ========================================================================================
+            string wktPrj = "GEOGCS['WGS 84',DATUM['WGS_1984',SPHEROID['WGS 84',6378137,298.257223563,AUTHORITY['EPSG','7030']],AUTHORITY['EPSG','6326']]," +
```
DGuidi, line 35: From what I understand, the projection string is used just ot check if the data is in a GEOGCS format. To me this check is useless, because we are not checking the geometry but a separate parameters. Maybe a check of the geometry values (inside the valid range of values) can be slower but useful?

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+        /// https://github.com/nettopologysuite/nettopologysuite/issues/129
+        /// https://github.com/DotSpatial/DotSpatial/blob/e30421d6ee5b0cdf2b1fa01f95f8aabc50c1997a/Source/DotSpatial.Data/Shape.cs#L611
+        /// https://github.com/DotSpatial/DotSpatial/issues/1029
+        /// https://github.com/NetTopologySuite/ProjNet4GeoAPI/wiki/Well-Known-Text
+        /// https://desktop.arcgis.com/en/arcmap/10.3/manage-data/editing-fundamentals/creating-and-editing-multipart-polygons.htm
+        /// https://desktop.arcgis.com/en/arcmap/10.3/guide-books/map-projections/what-happens-to-features-at-180-dateline-.htm
+        /// </remarks>
+        public static Geometry ToPolyExOp(List<Coordinate> inpCoords, string inpProj4Wkt, OgcGeometryType outType, double outTrimGap, double outDensifyResolution, bool isAttFixOutInvPolygons)
+        {
+            Debug.Print("$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$");
```
DGuidi, line 134: these "debug.print" should be removed from the main methods.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+
+            // ###################################################################################################################
+            // input coordinate checks
+            // ###################################################################################################################
+            if (inpCoords == null)
+            {
+                throw new Exception("Null input coordinates.");
```
DGuidi, line 141: please use "ArgomentNullException" instead of generic "Exception"

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+            }
+
+            // check the minimum number of input coordinates for an output geometry
+            switch (outType)
+            {
+                case OgcGeometryType.LineString:
+                    if (inpCoords.Count < 2)
+                        throw new Exception("Input list has fewer than 2 coordinates. To create a polyline, the minimum is 2 coordinates.");
```
DGuidi, line 149: ArgumentOutOfRange (of maybe InvalidOperationException?) here

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+                    break;
+                case OgcGeometryType.Polygon:
+                    if (inpCoords.Count < 3)
+                        throw new Exception("Input list has fewer than 3 coordinates. To create a polygon, the minimum is 3 coordinates.");
+                    break;
+                default:
+                    throw new Exception("Unsupported output geometry type.");
```
DGuidi, line 156: NotSupportedException if the operation can not be implemented, or NotImplementedException if you plan to extend the code to support these operations in the future

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+        /// https://github.com/NetTopologySuite/NetTopologySuite/blob/84aef3c6a6cd3c2bfa5e3034ba161555cef5a117/test/NetTopologySuite.Samples.Console/Operation/Polygonize/SplitPolygonExample.cs#L24
+        /// https://github.com/nettopologysuite/nettopologysuite/issues/129
+        /// https://github.com/DotSpatial/DotSpatial/blob/e30421d6ee5b0cdf2b1fa01f95f8aabc50c1997a/Source/DotSpatial.Data/Shape.cs#L611
+        /// https://github.com/DotSpatial/DotSpatial/issues/1029
+        /// https://github.com/NetTopologySuite/ProjNet4GeoAPI/wiki/Well-Known-Text
+        /// https://desktop.arcgis.com/en/arcmap/10.3/manage-data/editing-fundamentals/creating-and-editing-multipart-polygons.htm
+        /// https://desktop.arcgis.com/en/arcmap/10.3/guide-books/map-projections/what-happens-to-features-at-180-dateline-.htm
+        /// </remarks>
+        public static Geometry ToPolyExOp(List<Coordinate> inpCoords, string inpProj4Wkt, OgcGeometryType outType, double outTrimGap, double outDensifyResolution, bool isAttFixOutInvPolygons)
```
DGuidi, line 132: `IList` can be used, I think

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+
+                    // define the cutter polygon that is based on the dateline with a shift to put it in the same hemisphere as the shifted
+                    // polygon. Plus we also add the trim gap as the cutter polygon width.
+                    var cutterPgon = DatelineCutterPolygon(delNeg, delPos, shift, outTrimGap);
+                    DebugPrintCoords(cutterPgon.Coordinates.ToList(), "cutterPgon count");
+                    var valOp2 = new IsValidOp(cutterPgon);
+                    Debug.Print("cutterPgon isValid = {0}", valOp2.IsValid);
+
+                    // perform an Erase to split the polygon with a polygon. this consists of an Intersection and then a Difference.
+                    //var intx = pPolygonShifted.Boundary.Intersection(cutterPgon);
```
DGuidi, line 373: Old commented code that can be removed?

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+                var valOp = new IsValidOp(resGeoms);
+                Debug.Print("resGeoms isValid = {0}", valOp.IsValid);
+                if (!valOp.IsValid)
+                {
+                    // method 1
+                    resGeoms = resGeoms.Buffer(0);
+                    Debug.Print("resGeoms were fixed");
+
+                    // method 2
+                    //var pm = new PrecisionModel(10000000000.0);
```
DGuidi, line 487: Old code that can be removed? or it can be handled using a specific methofd parameter?

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi, line 132: "out" parameters (all the parameters following "inpProj4Wkt") can be moved to a single "parameter class", just for the sake of the readability

DGuidi:
I just added some "formal" comments about the code, I don't know so much about the algorithm but I think you can add few unit test methods to show how to use the class to obtain different kind of results.
thanks for the addition anyway.

euangordon:
@sindizzy This looks fantastic and fingers crossed it works well. 

One question: We have many polygons crossing the 180 meridian. With this change will we be able to intersect with those polygons?

I will try and test this out myself, just wondering if you know already

DGuidi:
> This looks fantastic and fingers crossed it works well

@euangordon can you suggest some more unit tests?

src/NetTopologySuite/NetTopologySuite.csproj
sindizzy: I wrote an article on my issue here https://github.com/NetTopologySuite/NetTopologySuite/issues/577
I didn't know how to fix it. It was suggested I remove a package and recompile. That seemed to work.

sindizzy:
> I just added some "formal" comments about the code, I don't know so much about the algorithm but I think you can add few unit test methods to show how to use the class to obtain different kind of results. thanks for the addition anyway.

Thanks I will spend some time this week to clean up the code and fix the commit authoring.

As for the tests, I added 3 tests on how to use the routine. I followed other routines in the same area ie test\NetTopologySuite.Samples.Console\Technique.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: I haven't done any unit testing in any of my projects so am a complete rookie at that. As you mentioned I just followed the pattern of testing in other routines in the Technique area.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: I suppose I can do that but say if someone has coordinates in UTM that go from 50 to 170 meters. That will pass the geographic check as we are only checking values and not units. My thought was that I didn't want users to just hand over their coordinates hoping it would work with any projection. The easiest way, for me at least, was to force them to provide a projection string.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Yes, thank you. Per my main write-up "The PR is also heavy on the debugs and I can remove those once the NTS devs can take a look at the initial proposal."

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Will do.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: I think I agree with your assessment here and will change my code. However, on my work related code I like to tell the user exactly what is wrong. At times I've encountered the ArgumentOutOfRange exception and wondered "ok what is the range then?".

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: got it. thanks.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: How different is using ListOf vs IList? I'm curious as to the pros and cons. 
Thanks.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: yep. will remove after the review is done. I'm an over-debugger I have to confess. I wanted to double check every output so  have tons of debugs. :)

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: well no. the routine to fix invalid geometries resulted from some research and eventually this article: https://github.com/NetTopologySuite/NetTopologySuite/issues/124
From that discussion it was noted that Buffer does not always work. So I added the second method in case I wanted to  add as an option later. I could probably make the input an Enum rather than a Boolean. So I would have 

```
    public enum attFixOutInvPolygons: int
    {
        FixNone= 0,
        FixViaBuffer=1,
        FixViaPrecision=2,
    }
```

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Question, is there a number of parameters where you move to a parameter class? Obviously if you have 30 params you don't want those all in the function call. If I have 3 or 4, I typically leave it in the function call as-is. Looking at the NTS core code there are various calls with 5 even 6 params right in the function call. So I'm wondering if this change is needed? I'm open to it.

src/NetTopologySuite/NetTopologySuite.csproj
DGuidi: My doubt is about pushing the package removal, I suggest to re-add the reference and then investigate the problem in #577

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi: I will help you with that

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi: I see your point, but maybe it's something related to the documentation of the method/api. my2cents

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi: using `IList`, the caller can use as input any implementation if the [IList interface](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ilist?view=net-6.0), like [arrays](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/)

DGuidi: Simple a suggestion, just because there are several "out" parameters, that _maybe_ can be grouped for the sake of the readability.
Totally unrelated, see something like JsonConvert.Deserialize(jsonString, jsonDeserializationParamsClass).

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi: you can leave the message. I see your point and ArgumentException can be used too. 
I must admit that, once you don't use the base `Exception` class, the proper exception to use can be arguable , so feel free to use the exception class you want. Often I use `InvalidOperationException`, but not sure I can be taken as an example :)

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi: I like the enum idea

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+            return resGeoms;
+        }
+
+        /// <param name="coordsInp">A list of coordinates that define a polyline or polygon.</param>
+        /// <param name="proj4WktInp">The projection of the input coordinates as a PROJ4 WKT string.</param>
+        /// <returns>
+        /// A Geometry class that contains the result of converting the coordinates into a Polyline or Polygon. The result could be multiple
+        /// geometries.
+        /// </returns>
+        private static Geometry ToPolyExOp(List<Coordinate> coordsInp, string proj4WktInp)
```
DGuidi, line 505: should be made public, I suppose

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi: @sindizzy  added the [unit test code](https://github.com/sindizzy/NetTopologySuite/commit/ed8b5368436a496200b94f3b7c2e6e6df2c86e75), hope to not have done a mess.
This is just a sample, please feel free to add your own asserts and other checks.

DGuidi:
not sure about how the unit tests in the XUniRunner are now failing, I see the same problem in my dev machine

src/NetTopologySuite/NetTopologySuite.csproj
FObermaier: This has to go away. If it does not work in your environment, you have to fix your environment.
If you want to turn it off temporarily, change `EnableApiCompat` to false (see 1st PropertyGroup)

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+            {
+                var resGeom = resGeoms.GetGeometryN(i);
+                string resWkt = wtr.Write(resGeom);
+                Debug.Print("geometry {0} = {1}", i, resWkt);
+                DebugPrintCoords(resGeom.Coordinates.ToList(), "resGeom geom [" + i.ToString() + "] count");
+            }
+        }
+
+        /// <summary>
+        /// This routine takes a list of Coordinates and attempts to convert them to a poly geometry (polyline or polygon). Special care is
```
FObermaier, line 90: What do I have to do when I have a `[Multi]LineString` that crosses the dateline?

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
FObermaier: I'd prefer `inpCoordinates` to be of type `CoordinateSequence`. Maybe this whole function should be encapsulated in a class implementing `IEntireCoordinateSequenceFilter` invoked by `Geometry.Apply(filter)`

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+                    // all other output types are not supported
+                    throw new Exception("Unsupported output geometry type.");
+            }
+
+            // ###################################################################################################################
+            // check the input projection info. currently only a geographic input projection is supported. we make many assumptions
+            // based on a geographic coordinate system. if the user wishes to use a projected coordinate system then they will have
+            // to reproject the points first and then use this process.
+            // ###################################################################################################################
+            if (inpProj4Wkt == null || !inpProj4Wkt.StartsWith("GEOGCS"))
```
FObermaier, line 181: This check (and the parameter) can be removed. One is only attempting to fix dateline issues when one has geographic coordinates. Maybe a test on the extent of the input coordinates as a replacement? In any case, if the width of the envelope is <180, there can't be a dateline crossing.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+            return _decimalDegrees < -180 ? (180 + (_decimalDegrees % 180)) : _decimalDegrees;
+        }
+
+        /// <summary>
+        /// Determines if a polygon crosses the dateline. This is slightly different than checking a LineString as the Polygon is
+        /// closed. So a segment exists between the last vertex and the first vertex.
+        /// </summary>
+        /// <param name="coordsInp">The input list of coordinates.</param>
+        /// <returns>True if the coordinates as a polygon cross the dateline, False otherwise.</returns>
+        private static bool IsPolygonCrossesDatelineEx(List<Coordinate> coordsInp)
```
FObermaier, line 537: Again, this should work on a `CoordinateSequence`

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
```diff
@@ -0,0 +1,802 @@
...
+            var cutterPolygon = new Polygon(pRing);
+
+            // we can densify the polygon but this seems to confuse the difference routine
+            // cutterPolygon = (Polygon)Densifier.Densify(cutterPolygon, 0.1); // densify to a specific resolution 0.1 degrees
+
+            // return the "new" dateline polygon that has been shifted and also has a gap from the dateline
+            return cutterPolygon;
+        }
+
+        private static void DebugPrintCoords(List<Coordinate> coords, string hdr)
```
FObermaier, line 764: Give these `Debug...` functions a return type of `string` use a `StringBuilder` to assemble the output and call the functions using `System.Diagnostics.Debug.WriteLine(~Debug~...)`

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
FObermaier: Or have an overload of the function taking a `Geometry` as an argument. In that case `IGeometryFilter` would be an option.

FObermaier (changes requested):
Instead of shifting all longitudes we could only shift the longitudes of geometries that cross the dateline by 360. In that case we have can use 2 constant intersection polygons and translate the geometries in the range >180 <=540 by -360

Just an idea.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
DGuidi: same suggestion I provided, see [reply](https://github.com/NetTopologySuite/NetTopologySuite/pull/580#discussion_r785487001)

src/NetTopologySuite/NetTopologySuite.csproj
sindizzy: Yes but I don't know how to go about fixing my environment. I'm not even sure where to start. I did a google search and not much came up. I'm just unfamiliar with ApiCompat and how it works.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Absolutely. I strived to add quite a bit of a description buttttttt there will be many that don't read it and just dive in. :)

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Sure thing. I am learning on the go.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Ok let me take a look at the NTS codebase and see how CoordinateSequence is used.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Im adding it as my next commit.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Ok I will take a loo. Thanks.

sindizzy:
> Instead of shifting all longitudes we could only shift the longitudes of geometries that cross the dateline by 360. In that case we have can use 2 constant intersection polygons and translate the geometries in the range >180 <=540 by -360

Yes I think you suggested that way back with my first post https://github.com/NetTopologySuite/NetTopologySuite/issues/330. I also have a todo in my function description. I think by the time I realized that may have been a better option I was already deep in the weeds with the current implementation. It shouldn't be too hard for me to do what you describe here and it actually might be a more logical flow.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Right now I only implemented polygon. That's what I so very needed and I wanted to make sure that was solid before I started experimenting with polylines.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: Users will no doubt try to use it with any projection and realize it doesn't work. I was trying to be proactive about feedback that it is broken when using UTM or other non-geographic projections. But as I said before, what if the UTM projection goes from 50 to 170 meters?

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: I will scope that out.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: I will probably take all that out here after Im done with the other suggestions. It was more for me than anything else. Thank you for the feedback.

sindizzy:
> Or have an overload of the function taking a Geometry as an argument. In that case IGeometryFilter would be an option.

Yes I think I can do that. Thanks.

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: So how would the Geometry as an input work?
1. The assumption that any geometry can be broken down to a coordinate list and then the user selects what kind of output comes from those coordinates? -OR-
2. The assumption that if I provide a Polygon then I want output as polygon. If I provide a polyline then I want polyline for output? What if the user inputs other non-supported geometry types?

What makes the most sense?

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: I think I can use something like:

`throw new ArgumentOutOfRangeException(nameof(count), count, "Must be non-negative.");`

test/NetTopologySuite.Samples.Console/Technique/SplitPolygonAtDateline.cs
sindizzy: This is a function internal to the process. I am converting the CoordnateSequence to a List<Coordinate> anyway since I require a bunch of List functionality.

DGuidi: I think you are using onlu `List.Exists` method, that can be re-implemented easily for coordinate sequences

sindizzy: I use more List functions than that I believe. If the CoordinateSequence does not provide IEnumerable then what are the alternatives?

sindizzy:
Ok I tried to change the commit authors to my personal account but I'm not sure that it worked. At this point I wonder if I should just try with another brand new PR.

FObermaier:
@sindizzy, don't worry about the CLA.

FObermaier:
Please have a look at https://gist.github.com/FObermaier/ba25a5b6ccc98cb86efbbd485d17bc01

sindizzy:
> Please have a look at https://gist.github.com/FObermaier/ba25a5b6ccc98cb86efbbd485d17bc01

What am I looking for?

sindizzy:
> Please have a look at https://gist.github.com/FObermaier/ba25a5b6ccc98cb86efbbd485d17bc01

Is that the final routine?

FObermaier:
@sindizzy, that is how I would do it, does it work for you?

sindizzy:
> @sindizzy, that is how I would do it, does it work for you?

Let me test it. Can I give you some feedback later in the week?

sindizzy:
Also on line 87 I am not certain I understand the need to union the results. if we have a multi part polygon and we split it along the dateline then the output may be something we don't expect after the union. I would have to give this some more thought.

FObermaier:
@sindizzy the example posted was not meant to replace your solution, it was meant to demonstrate how to work with `CoordinateSequence`s and `I[Entire]CoordinateSequenceFilter`s. I have next to no knowledge about dateline crossing caveats, everything I have to work with is planar or far enough from the dateline.

sindizzy:
Ahhhh gotcha. Ok let me use that as a base for my logic.

sindizzy:
@FObermaier Ok based on your suggestion I added a DatelineUtility class to my PR. I think I understand your preferred workflow now. 

I did change the main filter to
`private class IsCrossesDatelineFilter : IGeometryFilter`

This is because if a user passes a multi-part polygon then each polygon part could exist wholly in one hemisphere. If we use the CoordinateSequence then that could give us a false positive. Before I continue with this code I wanted to see if I am on the right track here. The filter cycles through each part in the geometry and then through the Coordinate array. I don't use the CoordinateSequence so wondering if that will pose a problem?

sindizzy:
Alright, I added a CoordinateSequence filter that is called from the Geometry filter. This should account for multi-part geometries as the input. If we **_just_** do a CoordinateSequence then the coordinate sweep could jump from one part to another and we could get a false positive.

test/NetTopologySuite.Samples.Console/Technique/DatelineUtility.cs
```diff
@@ -0,0 +1,233 @@
...
+            public bool GeometryChanged => false;
+
+            public bool IsCrossesDateline { get; private set; }
+
+            public void Filter(Geometry geometry)
+            {
+                // since we are filtering per geometry we could have multiple parts. so we must test each part individually for traversing
+                // dateline. if we were to test the CoordinateSequence with a IEntireCoordinateSequenceFilter filter then we could get false
+                // positives (ex: two islands on each side of the dateline).
+                System.Diagnostics.Debug.Print("num geoms={0}", geometry.NumGeometries);
```
FObermaier, line 130: I'd argue that if a single `Polygon` of a `MultiPolygon` crosses the dateline you should consider the whole geometry to cross the dateline.

test/NetTopologySuite.Samples.Console/Technique/DatelineUtility.cs
```diff
@@ -0,0 +1,233 @@
...
+                    //        currX > -180d && wasLT)
+                    //    {
+                    //        IsCrossingDateline = true;
+                    //        break;
+                    //    }
+                    //    if (currX < -180d) wasLT = true;
+                    //    else if (currX > -180d) wasGT = true;
+                    //}
+
+                    bool hasLonsLt180 = part.Coordinates.Any(c => c.X < 180);
```
FObermaier: Don't work on `Geometry.Coordinates`, that is potentially a bottleneck. For this check you iterate over the whole array twice + you might need to build the array of coordinates twice. Use `IEntireCoordinateSequenceFilter` to do that check.

FObermaier:
You don't need to split up into seperate filters. `IEntireCoordinateSequenceFilter` runs on all `CoordinateSequence`s of a geometry's components, as long as `IsDone` is false.

Note: Only `Point` and `LineString` geometry classes have `CoordinateSequence` members. So if you have a `Polygon` the `IEntireCoordinateSequenceFilter` will be applied to the shell and holes (which are `LinearRing`s (a special form of a `LineString`)) individually. Same for all the `Multi...` geometries and `GeometryCollection`s.

sindizzy:
@FObermaier Ahhhhh I think I got it now. Thanks for taking time to teach me how this works. I went back to **_only_** the IEntireCoordinateSequenceFilter and did a commit. I think this is what you were talking about so when you get a chance please take a look.

sindizzy:
I noticed you updated your routine so I started over with DatelineUtility2. Added my logic and the geographic CRS list. I will need to study it further to understand certain parts.

airbreather:
So, I haven't really been following this closely at all, but I do want to at least point out that the term "dateline" probably shouldn't show up in the code.  I don't mind referring to it that way informally, but the term "antimeridian" is a more accurate term to use when referring to the thing that causes problems for NTS, since the International Date Line deviates from the true meridian in several places in order to avoid chopping countries in half.

src/NetTopologySuite/NetTopologySuite.csproj
airbreather: > we have a fix for this

Not sure about that one, since the NuGet.config file does actually exist in this project... the question is what's causing it to get ignored?

euangordon:
Hi All, what is the status of this? Is there any chance it will get merged in?

sindizzy:
I was working on it and I lost steam. I'll pick it up again soon.

euangordon:
> I was working on it and I lost steam. I'll pick it up again soon.

Thanks @sindizzy we did try and use your fork but didn't manage to get it working. Hopeing that once it is in the release build we will manage to get it working.

As a workaround, I am working on splitting shapes that cross the dateline, but it painful.

FObermaier:
> Hi All, what is the status of this? Is there any chance it will get merged in?

In the meantime you can try https://gist.github.com/FObermaier/ba25a5b6ccc98cb86efbbd485d17bc01

sindizzy:
> > I was working on it and I lost steam. I'll pick it up again soon.
> 
> Thanks @sindizzy we did try and use your fork but didn't manage to get it working. Hopeing that once it is in the release build we will manage to get it working.
> 
> As a workaround, I am working on splitting shapes that cross the dateline, but it painful.

Yes I know how painful it is. You can try the algorithm here https://github.com/NetTopologySuite/NetTopologySuite/discussions/496 That's where the discussion started.

sindizzy:
I had this as a side project and am going to see if I can get back to refactoring it. I've spent the last couple of years transitioning from VB to C#. I've also upgraded my own algorithm slightly.

FObermaier:
Closed with https://github.com/NetTopologySuite/NetTopologySuite/pull/794

PR #595, opened 2022-02-15
Cache results of Geometry.Area and Geometry.Length computation

FObermaier:
Upon accessing Geometry.Area and Geometry.Length the values are always computed. This is usually not a problem except for scenarios where these values are compared to between different geometries.

The values are now cached and only computed once, unless the GeometryChangedAction invalidates the cached results.

### DIFF
--- a/src/NetTopologySuite/Geometries/Geometry.cs
+++ b/src/NetTopologySuite/Geometries/Geometry.cs
@@ -214,6 +214,9 @@ namespace NetTopologySuite.Geometries
         /// </summary>
         private Envelope _envelope;
 
+        private double? _area;
+        private double? _length;
+
         /// <summary>
         /// Sets the ID of the Spatial Reference System used by the <c>Geometry</c>.
         /// </summary>
@@ -476,7 +479,16 @@ namespace NetTopologySuite.Geometries
         /// Others return 0.0
         /// </summary>
         /// <returns>The area of the Geometry.</returns>
-        public virtual double Area => 0.0;
+        public virtual double Area
+        {
+            get => _area ?? 0d;
+            protected set => _area = value;
+        }
+
+        /// <summary>
+        /// Gets a value indicating if the value for <see cref="Area"/> has been set.
+        /// </summary>
+        protected bool IsAreaSet => _area.HasValue;
 
         /// <summary>
         /// Returns the length of this <c>Geometry</c>.
@@ -486,7 +498,16 @@ namespace NetTopologySuite.Geometries
         /// Others return 0.0
         /// </summary>
         /// <returns>The length of the Geometry.</returns>
-        public virtual double Length => 0.0;
+        public virtual double Length
+        {
+            get => _length ?? 0d;
+            protected set => _length = value;
+        }
+
+        /// <summary>
+        /// Gets a value indicating if the value for <see cref="Length"/> has been set.
+        /// </summary>
+        protected bool IsLengthSet => _length.HasValue;
 
         /// <summary>
         /// Computes the centroid of this <c>Geometry</c>.
@@ -650,6 +671,8 @@ namespace NetTopologySuite.Geometries
         public void GeometryChangedAction()
         {
             _envelope = null;
+            _area = null;
+            _length = null;
         }
 
         /// <summary>
--- a/src/NetTopologySuite/Geometries/GeometryCollection.cs
+++ b/src/NetTopologySuite/Geometries/GeometryCollection.cs
@@ -241,10 +241,14 @@ namespace NetTopologySuite.Geometries
         {
             get
             {
-                double area = 0.0;
-                for (int i = 0; i < _geometries.Length; i++)
-                    area += _geometries[i].Area;
-                return area;
+                if (!IsAreaSet)
+                {
+                    double area = 0.0;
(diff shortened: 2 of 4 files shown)


FObermaier:
@DGuidi could you test if this PR has any effect on the performance of Shapefile polygon ring sorting mechanism?

DGuidi:
@FObermaier I give a try and I see that `Issue123Test.CascadedUnionError2` (in `nts.io.shapefile` repository) fails with the "cache" applied: the expected `TopologyException` is not thrown

DGuidi:
@FObermaier  I don't see any particular improvement, but probably my [performance test](https://github.com/NetTopologySuite/NetTopologySuite.IO.ShapeFile/pull/80/commits/398140abe4b27082a66a90da388ae92a0721fbd2) is very rough (500000 polygons, each one with two holes)

BTW I tested in RELEASE both develop and "cache" branch, with and withont the flag to enable the ring sorting
```
develop:
flag DISABLED: n='20' => ms='719'
flag DISABLED: n='20' => ms='637'
flag DISABLED: n='20' => ms='639'
flag DISABLED: n='100' => ms='612'
flag ENABLED: n='20' => ms='2512'
flag ENABLED: n='20' => ms='2419'
flag ENABLED: n='20' => ms='2791'
flag ENABLED: n='100' => ms='2376'

cache:
flag DISABLED: n='20' => ms='665'
flag DISABLED: n='20' => ms='638'
flag DISABLED: n='20' => ms='624'
flag DISABLED: n='100' => ms='636'
flag ENABLED: n='20' => ms='2979'
flag ENABLED: n='20' => ms='2578'
flag ENABLED: n='20' => ms='2388'
flag ENABLED: n='100' => ms='2219'
```

DGuidi (approved):

FObermaier:
If there is no performance benefit, then we should just leave it as is.

airbreather:
I'll look at this later; if I don't respond again in the next two days, then please ping me because I probably forgot.

> I don't see any particular improvement, but probably my [performance test](https://github.com/NetTopologySuite/NetTopologySuite.IO.ShapeFile/pull/80/commits/398140abe4b27082a66a90da388ae92a0721fbd2) is very rough (500000 polygons, each one with two holes)

I would expect the improvement to be most visible when there are **few** polygons with **many** points.

> BTW I tested in RELEASE both develop and "cache" branch, with and withont the flag to enable the ring sorting.
in each row, the same test is repeated N times and the result is the average of the total milliseconds elapsed for each X of the N runs

I recommend using [BenchmarkDotNet](https://benchmarkdotnet.org) for performance testing, in practically every case, especially as the .NET team invests more into the tiered JIT compilation runtime feature.

DGuidi:
> I would expect the improvement to be most visible when there are **few** (edit: but not **too** few) polygons with **many** points.

I will try do do another check
> I recommend using [BenchmarkDotNet](https://benchmarkdotnet.org) for performance testing, in practically every case, especially as the .NET team invests more into the tiered JIT compilation runtime feature.

yep, I agree. I just need some time to work on it.

FObermaier:
> looks we are faster now

I'm disappointed, thought it was going to be, thanks anyway!

DGuidi:
I just created a simple perf test with benchmarkdotnet that show these results, using the same type of geometries created before
**develop**
image
**patch**
image
I will play with this sample and maybe try to run the test in an environment other than my crappy laptop

FObermaier:
This PR only makes sense if geometries are considered being long living objects. `Length` and `Area` are computed at least once since there is no way of providing the values during object creation. This has bothered me for `EnvelopeInternal`, where the situation is similar, for a long time.

@DGuidi, @airbreather, what do you think?

DGuidi:
maybe we can use a flag (a static property, [again](https://github.com/NetTopologySuite/NetTopologySuite.IO.ShapeFile/pull/80)?) to enable this behavoir, and possibly similar behaviours? so we can have the chance to test this in production and see if there was actual improvements, or enable it only for special cases.

airbreather:
At this time, I am not convinced that this is needed.  The benchmarks don't show a big enough improvement, and this particular implementation seems error-prone because of how closely the subclasses have to work with the base class and use its extra property.

If it is needed, I would personally change `GeometryChangedAction` to be `virtual` so that `Polygon.Area` and `LineString.Length` are the only two places where we use a cache: realistically, if there is a significant opportunity for a cache to improve things, then it's because of those two spots, not the places where we add up their results.

I'd also use `Lazy<double>` for the cache in those two places, because that class automatically handles certain edge cases re: threading that I prefer not to have to think about (it's **probably** not worse this way than it is without the cache, but I can't prove it, and whenever I get to the "I can't prove it" part, I just pull out `Lazy<T>` because it's almost never worth proving it).

airbreather:
> `Lazy<double>`

it would look something like
```csharp
private Lazy<double> _length;

public Constructor()
{
    ResetLengthCache();
}

public override double Length => _length.Value;

public override void GeometryChangedAction()
{
    if (_length.IsValueCreated)
    {
        ResetLengthCache();
    }

    base.GeometryChangedAction();
}

private void ResetLengthCache()
{
    _length = new Lazy<double>(() => Algorithm.Length.OfLine(_points));
}
```

FObermaier:
We will leave it as is

PR #604, opened 2022-04-13
LineSequencer.Reverse changes input geometries

inforithmics:
### Prerequisites

### Description
Running the LineSequencer changes the Input Geometries, because the Reverse LineString Function of LineSequencer makes an  inplace Reverse of the Coordinates instead of copying them.

### DIFF
--- a/src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
+++ b/src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
@@ -480,7 +480,11 @@ namespace NetTopologySuite.Operation.Linemerge
 
         private static LineString Reverse(LineString line)
         {
-            var pts = line.Coordinates;
+            var coordinates = line.Coordinates;
+            var pts = new Coordinate[coordinates.Length];
+
+            // copy first the array or else it gets reversed in the original Line
+            Array.Copy(coordinates, pts, coordinates.Length);
             Array.Reverse(pts);
             var rev = line.Factory.CreateLineString(pts);
             rev.UserData = line.UserData; // Maintain UserData in reverse process
--- a/src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
+++ b/src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
@@ -480,12 +480,9 @@ namespace NetTopologySuite.Operation.Linemerge
 
         private static LineString Reverse(LineString line)
         {
-            var coordinates = line.Coordinates;
-            var pts = new Coordinate[coordinates.Length];
+            // Use a new Coordinate Array to avoiding reversing the Coordinates of the line
+            var pts = line.CoordinateSequence.Reversed().ToCoordinateArray();
 
-            // copy first the array or else it gets reversed in the original Line
-            Array.Copy(coordinates, pts, coordinates.Length);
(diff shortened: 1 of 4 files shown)


src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
```diff
@@ -480,7 +480,11 @@ private Geometry BuildSequencedGeometry(IEnumerable<IEnumerable<DirectedEdge>> s
 
         private static LineString Reverse(LineString line)
         {
-            var pts = line.Coordinates;
+            var coordinates = line.Coordinates;
```
FObermaier, line 483: If we'd copy anything, then we'd copy the `CoordinateSequence`.
Gee, why doesn't his whole function work on `CoordinateSeqeuence` in the first place?

test/NetTopologySuite.Samples.Console/Tests/Operation/Linemerge/LineSequencerTest.cs
```diff
@@ -200,6 +203,27 @@ public void BadLineSequence()
             RunIsSequenced(wkt, false);
         }
 
+        [Test]
```
FObermaier, line 206: Can this be moved to `NetTopologySuite.Tests.NUnit.Operation.LineMerge.LineMergerTest`?

inforithmics: Yes I'll move it there

inforithmics:
Should I change the function to use `CoordinateSequence` ?

FObermaier:
Yes, please change to work on `CoordinateSequence`.

src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
airbreather: In fact, `CoordinateSequence` has a `.Reverse()` method that is documented to always return a reversed copy of the original sequence:
```csharp
var rev = line.Factory.CreateLineString(line.CoordinateSequence.Reverse());
```

src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
```diff
@@ -480,8 +480,9 @@ private Geometry BuildSequencedGeometry(IEnumerable<IEnumerable<DirectedEdge>> s
 
         private static LineString Reverse(LineString line)
         {
-            var pts = line.Coordinates;
-            Array.Reverse(pts);
+            // Use a new Coordinate Array to avoiding reversing the Coordinates of the line
+            var pts = line.CoordinateSequence.Reversed().ToCoordinateArray();
```
airbreather, line 484: `ToCoordinateArray()` might make an extra copy, and then `CreateLineString(Coordinate[])` might make yet another extra one, both depending on what the geometry factory uses for its `CoordinateSequenceFactory`.

Suggestion is to use `CreateLineString(CoordinateSequence)` directly from the result of `line.CoordinateSequence.Reversed()`

inforithmics: Done

FObermaier (approved):

PR #605, opened 2022-04-13
Fixed Reversing of EdgeSequence in LineSequencer

inforithmics:
### Prerequisites

### Description
Without this the Added Unit Test Fails.

Input to LineSequencer is 

```
"LINESTRING (0 1, 0 0)",
"LINESTRING (0 1, 0 2)",
"LINESTRING (0 3, 0 2)",
```

expected Result:

```
"MULTILINESTRING ((0 3, 0 2), (0 2, 0 1), (0 1, 0 0))"
```

current Result (without Reverse Fix):

```
"MULTILINESTRING ((0 2, 0 3), (0 1, 0 2), (0 0, 0 1))"
```

The Order of the Lines is correct but they are reversed.

### DIFF
--- a/src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
+++ b/src/NetTopologySuite/Operation/Linemerge/LineSequencer.cs
@@ -480,9 +480,8 @@ namespace NetTopologySuite.Operation.Linemerge
 
         private static LineString Reverse(LineString line)
         {
-            var pts = line.Coordinates;
-            Array.Reverse(pts);
-            var rev = line.Factory.CreateLineString(pts);
+            // Use a new CoordinateSequence to avoiding reversing the Coordinates of the line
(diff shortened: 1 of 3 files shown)


inforithmics:
Maybe the result of the linesequencer is expted because it tries to preserve the order of the input sequence.

inforithmics:
Yes the Input sequence is not correctly noded, But I think the Reverse Function for EdgeSequence does not work correctly here. 
The wrong noding only surfaced the bad behavior.

inforithmics:
So I removed the unnoded test and added a noded test that displays the same behavior.

inforithmics:
No, https://github.com/NetTopologySuite/NetTopologySuite/pull/604 fixed only that the LineString direction was changed by LineSquencer, So this pull reqeusts are not related. (Except that they fix Bugs in LineSequencer)

FObermaier (approved):

PR #608, opened 2022-05-12
Line segment hashcode collision reduction

Skibisky:
### Description
Line Segment GetHashCode creates collisions.
Seems to be straight lines of same length that are 10 units away in 1 axis.

Added a very simple unit test.
Added a performance test which should identify if things go bad.

Could probably increase robustness by testing lines of varying length/offset/direction.

### DIFF
--- a/src/NetTopologySuite/Geometries/LineSegment.cs
+++ b/src/NetTopologySuite/Geometries/LineSegment.cs
@@ -709,19 +709,6 @@ namespace NetTopologySuite.Geometries
                 hash = hash * 29 + _p1.Y.GetHashCode( );
                 return hash;
             }
-
-            long bits0 = BitConverter.DoubleToInt64Bits(_p0.X);
-            bits0 ^= BitConverter.DoubleToInt64Bits(_p0.Y) * 31;
-            int hash0 = (((int)bits0) ^ ((int)(bits0 >> 32)));
-
-            long bits1 = BitConverter.DoubleToInt64Bits(_p1.X);
-            bits1 ^= BitConverter.DoubleToInt64Bits(_p1.Y) * 31;
-            int hash1 = (((int)bits1) ^ ((int)(bits1 >> 32)));
-
-            // XOR is supposed to be a good way to combine hashcodes
-            return hash0 ^ hash1;
-
-            // return base.GetHashCode();
         }
     }
 }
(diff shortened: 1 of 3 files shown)


DGuidi:
I would like to highlight two main issues with this pull request, that personally made me not comfortable this PR to be accepted

1. The current implementation is the same [JTS code uses](https://github.com/locationtech/jts/blob/149b3890780b42836cee64b0c99c1daecb085ba2/modules/core/src/main/java/org/locationtech/jts/geom/LineSegment.java#L643), so for personal feelings I'm against introduce code differences (but we can open an issue in JTS code, of course)
2. I'm not comfortable with _unchecked_ code: but I must admit that I'm not anymore aligned with .net intrinsics, so I'm not sure that unckecked code can introduce issues with different runtimes and/or target compilation: maybe unckecked is used in other parts of NTS codebase? If it's already used, of course this second point is not valid

Skibisky:
Thanks for the link, I saw some references to JTS around. 
I will see if the hash collisions also occur in the Java environment.
The usage of 'unchecked' here permits integer overflow so that the runtime doesn't throw an exception.

DGuidi:
What I mean is: using unchecked in a codebase allow an assembly to be used in different context, as example in xamarin forms targeting iOS?

airbreather:
> What I mean is: using unchecked in a codebase allow an assembly to be used in different contexts, as example in xamarin forms targeting iOS?

`unchecked` should be fine.  It's actually the default, but this pattern is often used in places where overflow is **expected** to happen, since sometimes people turn on checking.

I would personally fix this by doing something more like
```csharp
return (_p0.X, _p0.Y, _p1.X, _p1.Y).GetHashCode();
```
but this works too

Skibisky:
JTS have made changes to LineSegment hashCode, is the submission in an acceptable state to continue?

FObermaier (commented):
Remove dead code.

src/NetTopologySuite/Geometries/LineSegment.cs
```diff
@@ -700,6 +700,16 @@ public override string ToString()
...
+            {
+                int hash = 17;
+                hash = hash * 29 + _p0.X.GetHashCode( );
+                hash = hash * 29 + _p0.Y.GetHashCode( );
+                hash = hash * 29 + _p1.X.GetHashCode( );
+                hash = hash * 29 + _p1.Y.GetHashCode( );
+                return hash;
+            }
+
             long bits0 = BitConverter.DoubleToInt64Bits(_p0.X);
```
FObermaier, line 713: Remove dead code

FObermaier (changes requested):

FObermaier (approved):

FObermaier:
Thanks for your contribution

PR #618, opened 2022-06-13
Fix incorrect line segment orientation

daniellivesey90:
### Prerequisites

### Description
Currently, in cases where the input line segment's start/end point is collinear with "this" line segment and its end/start point is to the right, an orientation of 0 (Collinear) is returned by the OrientationIndex(LineSegment) function.  This is inconsistent with the analogous left case, which returns an orientation index of 1 (Left).  Judging by the function's documentation, the function should return  -1 (Right) in the right case, as enforced by this PR.

### DIFF
--- a/src/NetTopologySuite/Geometries/LineSegment.cs
+++ b/src/NetTopologySuite/Geometries/LineSegment.cs
@@ -168,7 +168,7 @@ namespace NetTopologySuite.Geometries
                 return Math.Max(orient0, orient1);
             // this handles the case where the points are R or collinear
             if (orient0 <= 0 && orient1 <= 0)
-                return Math.Max(orient0, orient1);
+                return Math.Min(orient0, orient1);
             // points lie on opposite sides ==> indeterminate orientation
             return 0;
         }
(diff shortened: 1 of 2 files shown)


DGuidi (commented):
[Java code](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/LineSegment.java#L162) matches NTS implementation, so I would like to understand first if there's ab bug in JTS code, before fixing NTS implementation

daniellivesey90:
> [Java code](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/LineSegment.java#L162) matches NTS implementation, so I would like to understand first if there's ab bug in JTS code, before fixing NTS implementation

Can confirm that a bug in the JTS code and have submitted an issue here: https://github.com/locationtech/jts/issues/893

peetw:
The fix has now been merged upstream in JTS: https://github.com/locationtech/jts/pull/914

DGuidi:
merged, thanks.

FObermaier:
@DGuidi, please use "Squash and Merge" for bug-fixes in the future.

DGuidi:
yep, sorry about that.

PR #672, opened 2023-02-25
add .netstandard 2.1

inforithmics:
### Prerequisites

### Description
This adds .Net Standard 2.1 Target so that the System.Memory Dependency can be removed for this target. Because this causes often issues with Xamarin.iOS. (Which supports .Net Standard 2.1)

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -28,7 +28,7 @@ jobs:
       shell: bash # defaults disagree on how to quote the filter string
 
     - name: Pack
-      run: dotnet pack -c Release --no-build -o artifacts -p:NoWarn=NU5105
+      run: dotnet pack -c Release --no-build -o artifacts -p:NoWarn=NU5105 NetTopologySuite/NetTopologySuite.csproj
 
     - name: Upload
       uses: actions/upload-artifact@v2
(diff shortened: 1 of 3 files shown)


inforithmics:
1) Added netstandard 2.1 Target
2) Added Ignore for Jetbrains Rider
3) fixed build on .NET 7.0.200 SDK

airbreather:
Can you please elaborate more on the issues with the System.Memory package on Xamarin.iOS and confirm that they are not fixed in version 4.5.5 of that package?  Perhaps a link to a bug report or other write-up that shows what we're working around?

From my perspective, multi-targeting adds hidden complexity and risk, so I'm not very eager to start doing it again unless I can judge that the problems it solves are worse than the ones that it's likely to cause... and right now, I don't have a good grasp on what problems it solves.

inforithmics:
Concerning the System.Memory and iOS I found a different workaround. Adding a msbuild Target which removes it on iOS.

https://github.com/Mapsui/Mapsui/pull/1856/commits/b5005d23b2d0dd62a70080beacc6ee438c10deed

So the main change of this pull request is adding a .NET Standard 2.1 Target and therefore System.Memory does not need to be included as a dependency.

PR #676, opened 2023-03-15
Fix coverage simplifier issues

FObermaier:
JTS' commits
* locationtech/jts@e730a447314570f0b15f05c8f95620e559f9578f
* locationtech/jts@3203a931a72a4116097277e6b004c3aa8791da1e


FObermaier:
@airbreather this PR has issues with the difference in the `PriorityQueue`s used in NTS and JTS. NTS' is stable, the one used in JTS _unfortunately_ not. What do you think about how to solve this issue?

airbreather:
Does the issue go away after we pull in locationtech/jts@88d5965?

In case we still need `RemoveIf` in NTS, read on... I **just** realized that JTS no longer calls `removeIf` anymore.

------

I don't like this `RemoveIf` implementation... I would prefer something that doesn't need to keep re-heapifying for every item removed.  I **think** this is how to do it in `AlternativePriorityQueue<TPriority, TData>`?  I'm not positive (especially in the final `HeapifyDown` loop), I just threw it together somewhat hastily, but it does hit the same three failures.
```csharp
public int RemoveIf(Func<PriorityQueueNode<TPriority, TData>, bool> predicate)
{
    _ = predicate ?? throw new ArgumentNullException(nameof(predicate));

    int newCount = 0;
    for (int i = 0; i < this.Count; i++)
    {
        if (!predicate(this.nodes[i + 1]))
        {
            ++newCount;
            this.nodes[newCount] = this.nodes[i + 1];
            this.nodes[newCount].QueueIndex = newCount;
        }
    }

    int removed = this.Count - newCount;
    if (removed == 0)
    {
        return 0;
    }

    this.Count = newCount;
    for (int i = newCount / 2; i > 0; i--)
    {
        this.HeapifyDown(this.nodes[i]);
    }

    return removed;
}
```
And then `PriorityQueue<T>` can invoke it like:
```csharp
public int RemoveIf(Func<T, bool> predicate)
{
    return _queue.RemoveIf(node => predicate(node.Data));
}
```

airbreather:
Hm, just removing that line on our side (and doing nothing else) is not enough to make the tests start passing.  I haven't looked too much into this yet (still haven't), I'm trying to respond quickly before I vanish again.

At a glance, the test failure that concerns me the most is `TestTouchingHoles`, because it looks like JTS is deleting more coordinates than what NTS deletes.  The other two seem to delete the same number of coordinates, but **different** ones, which is perfectly understandable in the context of how these test-case polygons were probably built.

I would need to read more before saying anything definitive, but having engaged with this problem in the past, I imagine that if JTS is correct to delete more coordinates than what NTS deletes, then NTS is **potentially** doing something wrong.  But also, **that's not a certainty**: if it's implemented the way that I would have done it without having JTS as a reference (and a quick scan through just the class and method names suggests that I'm on the right track with this), then it is entirely possible that NTS just got unlucky when choosing which vertex to delete among multiple equally valid candidates.

If NTS is just unlucky, and we made no other mistake when porting this, then there's realistically nothing we can (or should) do other than update the test cases after confirming that this is really just bad luck.  One possible way to confirm this might be to try to find a test case where JTS is the unlucky one and NTS is lucky (which we can probably do just by scrolling the coordinates in the polygon rings?).

FObermaier:
I have ported NTS' `AlternativePriorityQueue<TPriority, TData>` and related `PriorityQueueNode<TPriority, TData>` class to java and used it within JTS. Both implementations yield the same results now.

src/NetTopologySuite/Utilities/PriorityQueue.cs
```diff
@@ -87,6 +88,15 @@ public T Peek()
                 : node.Data;
         }
 
+        public bool RemoveIf(Predicate<T> predicate)
+        {
+            var items = _queue.Where(t => t != null ? predicate(t.Data) : false) ?? Array.Empty<PriorityQueueNode<T, T>>();
```
airbreather, line 93: If you don't want to either defer adding `RemoveIf` by pulling in locationtech/jts@88d5965 early or else work on that other version of `RemoveIf` that I had in the comments, then we should at least do something like:
```suggestion
            var items = _queue.Where(t => t != null ? predicate(t.Data) : false).ToArray() ?? Array.Empty<PriorityQueueNode<T, T>>();
```
It's possible that the one caller will never care, and I have no idea why this isn't throwing any exceptions as-written, but nodes are getting missed this way.  The following test fails unless I either add the `.ToArray()` call above, or rewrite it the way I had in my non-review comment on the PR:
```csharp
[Test]
public void TestOrder2()
{
    var q = new PriorityQueue<int>();
    q.Add(1);
    q.Add(10);
    q.Add(5);
    q.Add(8);
    q.Add(-1);
    q.RemoveIf(_ => true);
    Assert.That(q.IsEmpty);
}
```

airbreather (changes requested):
I'm focusing just on the parts directly around the priority queue stuff, because I think that's the only reason why I was mentioned here, so I don't want to overstep.

src/NetTopologySuite/Utilities/PriorityQueue.cs
FObermaier: Having `RemoveIf` in place or not does not have any effect on the outcome.

> I have ported NTS' `AlternativePriorityQueue<TPriority, TData>` and related `PriorityQueueNode<TPriority, TData>` class to java and used it within JTS. Both implementations yield the same results now.

I had to use the following tweaks on the `Corner` class (both NTS and JTS) to __ensure__ stable ordering within the queue:
```patch
 src/NetTopologySuite/Simplify/Corner.cs | 9 ++++++++-
 1 file changed, 8 insertions(+), 1 deletion(-)

diff --git a/src/NetTopologySuite/Simplify/Corner.cs b/src/NetTopologySuite/Simplify/Corner.cs
index c1e5f6ad..cfc27283 100644
--- a/src/NetTopologySuite/Simplify/Corner.cs
+++ b/src/NetTopologySuite/Simplify/Corner.cs
@@ -6,11 +6,14 @@ namespace NetTopologySuite.Simplify
     internal sealed class Corner : IComparable<Corner>
     {
 
+        private static volatile long Counter = long.MinValue;
+
         private readonly LinkedLine _edge;
         private readonly int _index;
         private int _prev;
         private int _next;
         private double _area;
+        private readonly long _itemNo;
 
         public Corner(LinkedLine edge, int i)
         {
@@ -19,6 +22,8 @@ public Corner(LinkedLine edge, int i)
             _prev = edge.Prev(i);
             _next = edge.Next(i);
             _area = AreaFn(edge, i);
+
+            _itemNo = Counter++; ;
         }
 
         public bool IsVertex(int index)
@@ -55,7 +60,9 @@ private static double AreaFn(LinkedLine edge, int index)
          */
         public int CompareTo(Corner o)
         {
-            return _area.CompareTo(o._area);
+            int res = _area.CompareTo(o._area);
+            if (res != 0) return res;
+            return _itemNo.CompareTo(o._itemNo);
         }
 
         public Envelope Envelope

```

airbreather (approved):
Based on what I've seen so far and what's been discussed in the comments elsewhere on the PR, I'm perfectly happy with this, after changing "expected" values in the test cases as needed to make the tests pass.

PR #686, opened 2023-05-12
Fix/gh685

### LINKED ISSUE #685: Handle srsDimension attribute in GMLReader smarter
Hi,

NTS only deals with srsDimension when the attribute is on the posList inside a polygon. The use case I have is that the srsDimension attribute is on the Polygon (parent node) next to the srsName.

It would be nice if either way would work, but to make that happen I think the GeometryFactory should be extended. This way, when the Polygon gets parsed, it will set the Dimension on that factory, and next when any inner objects are being parsed, those factories will inherit from the Polygon factory and thus default to have that dimension set to whatever that parent node had in its srsDimension.

Quite doalble I'd say, but touching the GeometryFactory potentionally is also quite impactfull. It could be extending (non-breaking) the codebase, but still.

Is it worth doing this, or is this never going to make it into an accepted Pull Request in the first place?

Here is some sample code (notice how srsDimension is on the Polygon, this fails, but when on the posList this runs fine):
```C#
    [Fact]
    public void ParseSrsTest()
    {
        // Parse GML data
        string gmlData = "<gml:Polygon xmlns:gml=\"http://www.opengis.net/gml\" srsDimension=\"3\" srsName=\"http://www.opengis.net/gml/srs/epsg.xml#4326\" >" +
                 "<gml:exterior><gml:LinearRing><gml:posList>244420.388 488213.902 0.0 244407.058 488204.758 0.0 244413.487 488195.387 0.0 244418.813 488199.041 0.0 244415.419 488203.989 0.0 244423.422 488209.478 0.0 244420.388 488213.902 0.0</gml:posList></gml:LinearRing></gml:exterior>" +
                 "</gml:Polygon>";
        var reader = new GMLReader();
        var geom = reader.Read(gmlData);

        // Get the coordinates of the polygon
        var coords = ((Polygon)geom).ExteriorRing.Coordinates;

        // Print the coordinates
        foreach (var coord in coords)
        {
            Console.WriteLine(coord);
        }
    }
```

### Prerequisites

- [ ] I have written a descriptive issue title
- [ ] I have verified that I am running the latest version of NetTopologySuite
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [ ] I have searched [open](https://github.com/NetTopologySuite/NetTopologySuite/issues) and [closed](https://github.com/NetTopologySuite/NetTopologySuite/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
_(A description of the bug or feature)_

### Steps to Reproduce
_(List of steps, sample code, failing test or link to a project that reproduces the behavior)_

### System Configuration
Tell us about the environment where you are experiencing the bug:
- NetTopologySuite version:
- Other NetTopologySuite packages and versions:
- Environment (Operating system, version and so on):
- .NET Framework version:
- Additional information:

_Thanks for reporting the issue to NetTopologySuite!_

FObermaier:
Closes #685

### DIFF
--- a/src/NetTopologySuite/IO/GML2/GMLReader.cs
+++ b/src/NetTopologySuite/IO/GML2/GMLReader.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Text.RegularExpressions;
 using System.Xml;
 using System.Xml.Linq;
+using NetTopologySuite.Algorithm;
 using NetTopologySuite.Geometries;
 using NetTopologySuite.Utilities;
 
@@ -95,6 +96,12 @@ namespace NetTopologySuite.IO.GML2
             return Read(reader);
         }
 
+        protected static int ReadSrsDimension(XmlReader reader, int srsDimension)
+        {
+            string tmp = reader.GetAttribute("srsDimension");
+            return tmp == null ? srsDimension : XmlConvert.ToInt32(tmp);
+        }
+
         /// <summary>
         /// Reads the coordinate.
         /// </summary>
@@ -144,8 +151,8 @@ namespace NetTopologySuite.IO.GML2
         /// <summary>
         /// Extract a <see cref="Coordinate" /> from a pos entity string value.
         /// </summary>
-        /// <param name="value"></param>
-        /// <returns></returns>
+        /// <param name="value">An array of string ordinate values</param>
+        /// <returns>A coordinate</returns>
         protected Coordinate ReadPosAsCoordinate(string[] value)
         {
             double[] ordinates = new double[Math.Min(3, value.Length)];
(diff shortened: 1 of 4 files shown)


src/NetTopologySuite/IO/GML2/GMLReader.cs
```diff
@@ -177,14 +185,68 @@ protected IEnumerable<Coordinate> ReadPosListAsCoordinates(int numOrdinates, str
...
+                int seperatorIndex = coordinatesSpan.IndexOf(' ');
+                if (seperatorIndex == -1)
+                    seperatorIndex = coordinatesSpan.Length;
+
+                var ordinateSpan = coordinatesSpan.Slice(0, seperatorIndex);
+                if (!ordinateSpan.IsEmpty)
+                {
+#if NET7
+                    if (double.TryParse(ordinateSpan, System.Globalization.NumberStyles.Number,
+                        System.Globalization.NumberFormatInfo.InvariantInfo, out double ordinate))
```
airbreather, line 210: This part only does anything if we go back to multi-targeting.  Though it's not actually **wrong**, so I'm inclined to say let's keep it around.

Maybe `#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER`, instead of `NET7`, since those are the minimum versions that we need for this API?

src/NetTopologySuite/IO/GML2/GMLReader.cs
```diff
@@ -161,14 +168,15 @@ protected Coordinate ReadPosAsCoordinate(string[] value)
         /// <summary>
         /// Extract a <see cref="Coordinate" /> from a x,y string value.
         /// </summary>
+        [Obsolete]
```
airbreather, line 171: Suggestion: I personally always try to add messages indicating what current callers should use instead.  If we're not **actually** expecting people to inherit from this class, then it might be a decent idea to add this message to all the `protected` members (though that might also be overkill):
```suggestion
        [Obsolete("This class should generally not be inherited from.  Protected members do not necessarily have a perfectly stable API.")]
```

src/NetTopologySuite/IO/GML2/GMLReader.cs
```diff
@@ -177,14 +185,68 @@ protected IEnumerable<Coordinate> ReadPosListAsCoordinates(int numOrdinates, str
...
 
+        protected CoordinateSequence ReadPosList(XmlReader reader, int srsDimension, CoordinateSequenceFactory csFactory)
+        {
+            Assert.IsTrue(IsStartElement(reader, "posList"));
+
+            // Read, update srsDimension
+            srsDimension = ReadSrsDimension(reader, srsDimension);
+            reader.Read();
+
+            var coordinatesSpan = reader.ReadContentAsString().Trim().AsSpan();
```
airbreather, line 196: :question: (nit) I haven't pulled this up yet, but does `.AsSpan().Trim()` work with this version of the API?  If so, that's a free small improvement, since it saves an allocation without any downsides.

airbreather (commented):
I'm going to assume that the logic is correct, because I don't think it's worth me pulling up the spec to validate that part of it.

For my own reference: I was concerned when I saw that the former `[Obsolete]` warning was resolved in a different way than what the message said would happen, but then I saw that the `[Obsolete]` warning was added **after** the previous release, so the only people who would be impacted are the people who have started using the prerelease packages since then, so that's not an issue at all.

Issue #698, opened 2023-08-07
Issue with Error Handling and Width Calculation in VariableBuffer Class


Alison-97:
### Description
We have identified an issue related to the VariableBuffer class from the NetTopologySuite library. When one of the widths in the array provided to the Buffer method is set to 0, it results in an unexpected error. The error message is as follows:
`NetTopologySuite.Geometries.TopologyException: 'side location conflict: arg 1 [ (2216873.0320880944, 5312576.81797011) ]'`

### Steps to Reproduce
To reproduce the issue, we have prepared a code snippet that demonstrates the problem:
The following code is working fine:

```
using NetTopologySuite.Geometries;
using NetTopologySuite.Operation.Buffer;

namespace tryNetTopo
{
    internal class Program
    {
        static void Main(string[] args)
        {
            var polylineCoordinates = new Coordinate[]
              {
                    new Coordinate(2216883.9690149571, 5312555.8022460323),
                    new Coordinate(2216880.7776965303, 5312561.9943400519),
                    new Coordinate(2216874.5057125678, 5312574.1638290221),
                    new Coordinate(2216873.0320880935, 5312576.8179701111),
                    new Coordinate(2216872.0585015123, 5312578.3818236338),
                    new Coordinate(2216871.4142499096, 5312579.3486044565)
              };

            double[] arr = new double[polylineCoordinates.Length];
            for (int i = 0; i < arr.Length; i++)
            {
                arr[i] = 1;
            }

            arr[3] = 0;

            var line = new LineString(polylineCoordinates);

            var result = VariableBuffer.Buffer(line, arr);
        }
    }
}
```

The following code contains one extra coordinate, and the error will be thrown:
```
using NetTopologySuite.Geometries;
using NetTopologySuite.Operation.Buffer;

namespace tryNetTopo
{
    internal class Program
    {
        static void Main(string[] args)
        {
            var polylineCoordinates = new Coordinate[]
              {
                    new Coordinate(2216883.9690149571, 5312555.8022460323),
                    new Coordinate(2216880.7776965303, 5312561.9943400519),
                    new Coordinate(2216874.5057125678, 5312574.1638290221),
                    new Coordinate(2216873.0320880935, 5312576.8179701111),
                    new Coordinate(2216872.0585015123, 5312578.3818236338),
                    new Coordinate(2216871.4142499096, 5312579.3486044565),
[...]

PR #700, opened 2023-08-18
Fix VariableBuffer to handle zero distances

Alison-97:
### Description

This patch is adopted from the fix in JTS - [Fix VariableBuffer to handle zero distances](https://github.com/locationtech/jts/pull/997)

It seeks to resolve an [issue ](https://github.com/NetTopologySuite/NetTopologySuite/issues/698) within the VariableBuffer class of the NTS library, which was previously identified and successfully addressed in the JTS library.


src/NetTopologySuite/Operation/Buffer/VariableBuffer.cs
```diff
@@ -8,8 +8,9 @@ namespace NetTopologySuite.Operation.Buffer
     /// <summary>
     /// Creates a buffer polygon with a varying buffer distance
     /// at each vertex along a line.
+    /// Vertex distances may be zero.
     /// <para/>
-    /// Only single lines are supported as input, since buffer widths
+    /// Only single linestring are supported as input, since buffer widths
```
FObermaier, line 13: linestrings

FObermaier (changes requested):
Thanks for you valuable contribution. Would you be so kind to port the changes to the unit test as well? Thanks

Alison-97:
> Thanks for you valuable contribution. Would you be so kind to port the changes to the unit test as well? Thanks

The unit test has been ported in

FObermaier (approved):

PR #703, opened 2023-09-02
Make GeometryCollection implement IReadOnlyList

### LINKED ISSUE #702: Should GeometryCollection implement IReadOnlyList<Geometry>?
GeometryCollection currently implements `IEnumerable<Geometry>`, but not `IList<Geometry>`. However, it exposes a very IList-like interface (GetGeometryN accepting an array index, Geometries returning the internal array). So it could easily implement `IReadOnlyList<Geometry>`, which basically exposes Count and an indexer, making it easier to use GeometryCollection in some standard .NET contexts.

For example, some LINQ operators have special behavior when a type implement IList/IReadOnlyList (in EF we have some special behavior too). This would simply kick in if GeometryCollection implemented that.

/cc @ajcvickers

roji:
Closes #702


FObermaier (approved):
Thanks

PR #712, opened 2023-09-15
Fix point's ordinate properties to not throw exceptions

FObermaier:
`Point`'s X and Y properties throw `ArgumentOutOfRangeException` when accessing or setting on an empty point.
This is undesireable hence the semantics are changed to:
* return double.NaN on the getter
* do nothing on the setter

### DIFF
--- a/src/NetTopologySuite/Geometries/Point.cs
+++ b/src/NetTopologySuite/Geometries/Point.cs
@@ -68,7 +68,7 @@ namespace NetTopologySuite.Geometries
             if (coordinates == null)
                 coordinates = factory.CoordinateSequenceFactory.Create(new Coordinate[] { });
             NetTopologySuite.Utilities.Assert.IsTrue(coordinates.Count <= 1);
-            this._coordinates = coordinates;
+            _coordinates = coordinates;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/NetTopologySuite/Geometries/Point.cs
```diff
@@ -113,46 +113,44 @@ public override double[] GetOrdinates(Ordinate ordinate)
...
+        /// when this property is accessed or set
+        /// </remarks>
         public double X
         {
             get
             {
-                if (Coordinate == null)
-                    throw new ArgumentOutOfRangeException("X called on empty Point");
-                return Coordinate.X;
+                return !IsEmpty ? CoordinateSequence.GetX(0) : double.NaN;
```
airbreather, line 126:
```suggestion
                return !IsEmpty ? CoordinateSequence.GetX(0) : Coordinate.NullOrdinate;
```

src/NetTopologySuite/Geometries/Point.cs
```diff
@@ -113,46 +113,44 @@ public override double[] GetOrdinates(Ordinate ordinate)
...
+        /// when this property is accessed or set
+        /// </remarks>
         public double Y
         {
             get
             {
-                if (Coordinate == null)
-                    throw new ArgumentOutOfRangeException("Y called on empty Point");
-                return Coordinate.Y;
+                return !IsEmpty ? CoordinateSequence.GetY(0) : double.NaN;
```
airbreather, line 145:
```suggestion
                return !IsEmpty ? CoordinateSequence.GetY(0) : Coordinate.NullOrdinate;
```

airbreather (approved):
Should the `double.NaN`s be `Coordinate.NullOrdinate` instead?  Same value, just a different name...

I'm fine either way.

src/NetTopologySuite/Geometries/Point.cs
FObermaier: Good catch!

PR #714, opened 2023-09-25
Fix LinearRing Type on Polygon Class

Jeongyong-park:
### Prerequisites

### Description

It was LinearRing Type not  LineString Type in  [Polygon Class of JTS](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Polygon.java)

### DIFF
--- a/src/NetTopologySuite/Geometries/Polygon.cs
+++ b/src/NetTopologySuite/Geometries/Polygon.cs
@@ -251,7 +251,7 @@ namespace NetTopologySuite.Geometries
         /// <summary>
         ///
         /// </summary>
-        public LineString ExteriorRing => _shell;
+        public LinearRing ExteriorRing => _shell;
 
         /// <summary>
         ///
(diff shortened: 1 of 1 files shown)


Jeongyong-park:
@dotnet-policy-service agree

airbreather:
> It was LinearRing Type not LineString Type in [Polygon Class of JTS](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Polygon.java?rgh-link-date=2023-09-25T07%3A00%3A13Z)

Well, [it **was** `LineString`](https://github.com/locationtech/jts/blob/jts-1.16.1/modules/core/src/main/java/org/locationtech/jts/geom/Polygon.java) until locationtech/jts#290.

This would be a breaking change, one which I would not be inclined to take because callers can always use:
- `Shell` instead of `ExteriorRing`
- `Holes` instead of `InteriorRings`
- `Holes[n]` instead of `GetInteriorRingN(n)`

FObermaier:
As @airbreather already said, if you need `LinearRing` now use `Shell` and `Holes[idx]` property.
This change has been omitted intentionally to avoid a breaking change.

I'd postpone this to the point where we consider doing a new major release.

PR #748, opened 2024-04-10
Allow extensibility without breaking compat.

airbreather:
This should give us a reasonable way to give all the same flexibility as JTS without breaking compatibility or creating risky situations by taking advantage of multi-targeting for this.

### DIFF
--- a/src/NetTopologySuite/Noding/ISegmentString.cs
+++ b/src/NetTopologySuite/Noding/ISegmentString.cs
@@ -19,14 +19,6 @@ namespace NetTopologySuite.Noding
         /// </summary>
         int Count { get; }
 
-//#if NETSTANDARD2_1_OR_GREATER
-//        /// <summary>
-//        /// Gets the segment string <c>Coordinate</c> at a given index
-//        /// </summary>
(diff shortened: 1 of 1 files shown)


airbreather:
> This way we have a new interface that we don't use/implement internally. What is the point in doing that?

Without knowing any details about how the new methods are specifically being used (or intended to be used in the future), I just applied a pattern that brings NTS as close to JTS as possible, in the most compatible way possible. The interface is public, and it's exposed by plenty of other public APIs, so I figured that there had to be some reason why these were added to the interface in JTS with default implementations that library consumers could replace in their own implementations of these interfaces (as opposed to static helper methods that they cannot).

If you think that this is such an unlikely use case that we should just hold off and wait for someone to complain instead of preemptively going through all this trouble, then I'm perfectly happy with that and will close this PR immediately. I only stepped in because I was summoned to look at an area where there was commented-out code, and I knew of a pattern that could make a more perfect port of the JTS API.

FObermaier:
I really appreciate that you did look into that code.

I'll prepare `ISegmentString2` interface with added methods and leave it -as well as its uasge- commented out, 

Thanks

PR #751, opened 2024-04-15
No boxing in AbstractSTRtree loop

zgabi:
### Prerequisites
In my project there are a lot of unncessary boxing in the `AbstractSTRtree.QueryInternal` method.
image

It is because the `node.ChildBoundables` returns `IList<IBoundable<T, TItem>>`, so the GetEnumerator returns `IEnumerator<IBoundable<T, TItem>`. But in the behind it is `List<IBoundable<T, TItem>>`, so the actual type of the iterator is `List<IBoundable<T, TItem>>` which is a value type, so it is boxed in the foreach loop.
Foreach converted to for loop.

- [X] I have written a descriptive pull-request title
- [X] I have verified that there are no overlapping [pull-requests](https://github.com/NetTopologySuite/NetTopologySuite/pulls) open
- [X] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository.

### Description
_(A description of the changes proposed in the pull-request_)

_Thanks for contributing to NetTopologySuite!_

### DIFF
--- a/src/NetTopologySuite/Index/Strtree/AbstractSTRtree.cs
+++ b/src/NetTopologySuite/Index/Strtree/AbstractSTRtree.cs
@@ -314,8 +314,10 @@ namespace NetTopologySuite.Index.Strtree
 
         private void QueryInternal(T searchBounds, AbstractNode<T, TItem> node, IList<TItem> matches)
         {
-            foreach (var childBoundable in node.ChildBoundables)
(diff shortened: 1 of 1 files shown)


airbreather:
Thanks, good catch. I added some more fixes to hit everything that used `ChildBoundables` to make sure that other code paths get the same optimization.

PR #752, opened 2024-04-15
Add IsValid(int idx) and IsClosed to CordinateSequence

zgabi:
### Prerequisites

- [X] I have written a descriptive pull-request title
- [X] I have verified that there are no overlapping [pull-requests](https://github.com/NetTopologySuite/NetTopologySuite/pulls) open
- [X] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository.

### Description
During the validation of the LineString object some extra Coordiate objects are created when not the default CoordinateSequence is used. (For example a PackedCoordinateSequence or a custom sequence)
It would be doog if the CoordinateSequence could validate the Nth coordinate or check whether it is closed.
This feature request/PR adds 2 new virtual methods/property to the base CoordinateSequence class to check the validity of the Nth coordiante and whether the sequence is closed without getting the actual coordinate object.
It also adds an implementation for the PackedCoordinateSequence class.


src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
```diff
@@ -46,6 +46,32 @@ public sealed override Coordinate GetCoordinate(int i)
             return GetCoordinateCopy(i);
         }
 
+        /// <summary>
+        /// Returns whether the i'th coordinate is valid
+        /// </summary>
+        public virtual bool IsValidCoordinate(int i)
```
FObermaier, line 52: `public sealed override bool IsValidCoordinate(int i)`

src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
```diff
@@ -46,6 +46,32 @@ public sealed override Coordinate GetCoordinate(int i)
...
+        {
+            double d = GetX(i);
+            if (double.IsInfinity(d) || double.IsNaN(d)) return false;
```
FObermaier, line 55: `if (!double.IsFinite(d)) return false;`

src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
```diff
@@ -46,6 +46,32 @@ public sealed override Coordinate GetCoordinate(int i)
...
+
+            d = GetY(i);
+            if (double.IsInfinity(d) || double.IsNaN(d)) return false;
```
FObermaier, line 58: dito

FObermaier (changes requested):
I tend to support this enhancement.
I'm not sure about the naming of `IsValidCoordinate(int)`, I think `IsCoordinateValidAt(int)` is slightly better.

Unit tests are missing

src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
zgabi: It is not netstandard2.0.. .it does not compile.

zgabi:
IsValidCoordinate renamed to IsCoordinateValidAt.
IsValidCoordinate is sealed in the packed class.
And added unit test for the packed classes. The base class is covered by a lot of tests.

zgabi:
Why is this try-catch here?
https://github.com/NetTopologySuite/NetTopologySuite/blob/b2897e04b123744c87af4db4463674cffaf54ca9/test/NetTopologySuite.Tests.NUnit/Geometries/Implementation/PackedCoordinateSequenceTest.cs#L203-L210

I accicdentally copied this method for the new test methods... Now I removed the try-catches from my new test methods.

FObermaier:
> `double.IsFinite` is not available in netstandard2.0, however you have an aggressive inlined private method in the `Coordinate `class. Maybe this method could be moved to some helper class and use the same method from all locations. Maybe to the `Mathematics.MathUtils` class or to the `Utilities.Global` class.

I propose to add
```C#
/// <summary>
/// Tests if a value is a valid for an ordinate.
/// Valid ordinate values are finite.
/// </summary>
/// <param name="value">The ordinate value</param>
/// <returns><c>true</c> if <paramref name="value"/> has a finite value</returns>
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static bool IsValidOrdinateValue(double value)
{
#if NETSTANDARD2_1_OR_GREATER
    return double.IsFinite(value);
#else
    return (~BitConverter.DoubleToInt64Bits(value) & 0x7ff0000000000000L) != 0;
#endif
}

```
to either `Coordinate` or `Coordinates`-utility class

airbreather:
> > `double.IsFinite` is not available in netstandard2.0, however you have an aggressive inlined private method in the `Coordinate `class. Maybe this method could be moved to some helper class and use the same method from all locations. Maybe to the `Mathematics.MathUtils` class or to the `Utilities.Global` class.
> 
> I propose to add
> 
> ```cs
> /// <summary>
> /// Tests if a value is a valid for an ordinate.
> /// Valid ordinate values are finite.
> /// </summary>
> /// <param name="value">The ordinate value</param>
> /// <returns><c>true</c> if <paramref name="value"/> has a finite value</returns>
> [MethodImpl(MethodImplOptions.AggressiveInlining)]
> public static bool IsValidOrdinateValue(double value)
> {
> #if NETSTANDARD2_1_OR_GREATER
>     return double.IsFinite(value);
> #else
>     return (~BitConverter.DoubleToInt64Bits(value) & 0x7ff0000000000000L) != 0;
> #endif
> }
> ```
> 
> to either `Coordinate` or `Coordinates`-utility class

Having `IsValidOrdinateValue` sounds fine to me, but I'd just use the `BitConverter` version and not bother with the `NETSTANDARD2_1_OR_GREATER` check. If anything, based on a small benchmark, the `BitConverter` version appears like it might be faster than calling `double.IsFinite` directly (at least on my machine with `net8.0`).

FObermaier (approved):
Thanks for finding the other places where ordinate validity plays a role!

FObermaier:
I've thought a bit about `IsClosed` implemented on `CoordinateSequence`:   
We have the `IsClosed` property on `LineString` and `LinearRing`. Both have different semantics. For empty `LinearRing`s this leads to the situation that `ring.IsClosed != ring.CoordinateSequence.IsClosed` which is unfortunate.

zgabi:
Yes, emply `LineString` (base class) is not closed. Empty `LineRing` is cloed. But it was in this way earlier.
The implentation of the `CoordinateSequence.IsClosed` is the same as the previous `LineString` implementation.

What do you suggest?

zgabi:
What about this PR? Should I change anything else?
Currently the `IsClosed` value in this PR equivalent with the current "develop" code. So could you please merge it?
The `IsClosed` can be changed in a separate task/PR if you agree on how to change it.

bjornharrtell (approved):
LGTM

AsgerPetersen (approved):
I have taken the liberty to add a little documentation

PR #757, opened 2024-05-17
Compute CubicBezier Iterpolation Parameters once

rkolka:
### Prerequisites

### Description

Interpolation parameters are never changing (for a given number of vertices). ComputeIterpolationParameters can be called just once, ie `static`.

No change in behavior. 
(Although perhaps _numVerticesPerSegment is meant to be 17?)

### Discussion
If n is number of vertices per original segment then n-1 is the number of resulting segments. We might prefer the last number to be a power of 2 and _numVerticesPerSegment be 16 + 1.
If _numVerticesPerSegment-1 is a reasonably small power of 2 (eg 16 or 32) then fractions have finite representations in `double` binary and sum of parameters is exactly 1. With _numVerticesPerSegment = 16 there is a slight diff (less than 2E-16) sometimes. But even then we can make the sum exactly 1 easily. No need for division in CubicBezier.

### DIFF
--- a/src/NetTopologySuite/Shape/CubicBezierCurve.cs
+++ b/src/NetTopologySuite/Shape/CubicBezierCurve.cs
@@ -2,7 +2,6 @@
 using NetTopologySuite.Geometries;
 using NetTopologySuite.Geometries.Utilities;
 using System;
-using System.Diagnostics;
 
 namespace NetTopologySuite.Shape
 {
--- a/src/NetTopologySuite/Shape/CubicBezierCurve.cs
+++ b/src/NetTopologySuite/Shape/CubicBezierCurve.cs
@@ -2,6 +2,7 @@
 using NetTopologySuite.Geometries;
 using NetTopologySuite.Geometries.Utilities;
 using System;
+using System.Diagnostics;
 
 namespace NetTopologySuite.Shape
 {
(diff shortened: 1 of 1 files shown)


src/NetTopologySuite/Shape/CubicBezierCurve.cs
```diff
@@ -82,7 +82,7 @@ public static Geometry Create(Geometry geom, Geometry controlPoints)
         }
 
         private readonly double _minSegmentLength = 0.0;
-        private readonly int _numVerticesPerSegment = 16;
+        private static readonly int _numVerticesPerSegment = 16;
```
FObermaier, line 85: Can we expose this as a static property and compute the interpolation parameters when it is changed?
Otherwise `const` and `NumVerticesPerSegment`

src/NetTopologySuite/Shape/CubicBezierCurve.cs
```diff
@@ -92,7 +92,7 @@ public static Geometry Create(Geometry geom, Geometry controlPoints)
         private int _controlPointIndex;
   
         private Coordinate[] bezierCurvePts;
```
FObermaier, line 94: `_bezierCurvePts`

FObermaier (changes requested):
Good catch, thanks! I only have some minor change requests/suggestions.

bjornharrtell:
I wonder if this should be upstreamed first?

Issue #790, opened 2025-06-02
Buffer inconsistency with regards to v2.5.0


hahschaa:
Hi there,

I was migrating from version 2.5.0 to 2.6.0, but found that some things started to act differently. I've managed to narrow it down to a single method: `Geometry.Buffer(double distance, int quadrantSegments, EndCapStyle endCapStyle)`.

I've attached an example project that shows the different behavior in both versions:
[Nts2.6.0Buffer.zip](https://github.com/user-attachments/files/20556114/Nts2.6.0Buffer.zip)

I load a polygon from a file, then buffer it. In v2.5.0 the number of points increases as expected. In the newer version the number of points stays the same. This does not hold for all cases, e.g. a simple rectangle does work.

Thanks for your hard work, love the project.

With kind regards, 

Harm

FObermaier:
I assume this has to do with the issue described in [this JTS issue](https://github.com/locationtech/jts/issues/1131).
The fix would be to implement https://github.com/locationtech/jts/pull/1132.

hahschaa:
Thanks for your answer. Is there anything I can do to circumvent this? Or is the only option to keep using v2.5.0 for now. I'd contribute to the code if I could, but I'm afraid my knowledge of this subject is too limited.

michaldengusiak:
@FObermaier  We also have encountered an exception when using BufferCurveSetBuilder.GetCurves() with a valid polygon geometry and a negative buffer. This was previously working as expected in older versions.


```c sharp
// Input Polygon:
Geometry geometry = new WKTReader(new GeometryFactory())
    .Read("POLYGON ((0 0, 0 -7.439614, 12.134293 -7.439614, 12.134293 0, 0 0))");

// Validity checks:
bool valid = geometry.IsValid; // true
var  isValidOp = new IsValidOp(geometry); // ValidationError: null (geometry is valid)

// Precision model:
var precisionModel = new PrecisionModel(1 / 1E-6);

// Buffer parameters we used some not sure if best one:
var bufferParameters = new BufferParameters
{
    EndCapStyle = EndCapStyle.Square,
    JoinStyle = JoinStyle.Mitre
};

double offset = -0.1;

// This call throws an exception (worked in previous version):
var bufferCurveSetBuilder = new BufferCurveSetBuilder(geometry, offset, precisionModel, bufferParameters);
[...]

PR #793, opened 2025-07-14
Minor code cleanup

hahschaa:
### Prerequisites

### Description

I checked out the repo to see if I could make the changes required for [issue 790](https://github.com/NetTopologySuite/NetTopologySuite/issues/790), but saw some minor code cleanup that I could apply.

### DIFF
--- a/src/NetTopologySuite/Algorithm/MinimumBoundingCircle.cs
+++ b/src/NetTopologySuite/Algorithm/MinimumBoundingCircle.cs
@@ -267,7 +267,7 @@ namespace NetTopologySuite.Algorithm
             // handle degenerate or trivial cases
             if (_input.IsEmpty)
             {
-                _extremalPts = new Coordinate[0];
+                _extremalPts = Array.Empty<Coordinate>();
                 return;
             }
             if (_input.NumPoints == 1)
--- a/src/NetTopologySuite/Densify/Densifier.cs
+++ b/src/NetTopologySuite/Densify/Densifier.cs
@@ -159,7 +159,7 @@ namespace NetTopologySuite.Densify
                 // prevent creation of invalid LineStrings
                 if (parent is LineString && newPts.Length == 1)
                 {
-                    newPts = new Coordinate[0];
+                    newPts = Array.Empty<Coordinate>();
                 }
                 return Factory.CoordinateSequenceFactory.Create(newPts);
             }
--- a/src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequence.cs
+++ b/src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequence.cs
@@ -64,7 +64,7 @@ namespace NetTopologySuite.Geometries.Implementation
         {
             if (coordinates == null)
             {
-                Coordinates = new Coordinate[0];
+                Coordinates = Array.Empty<Coordinate>();
             }
             else
             {
@@ -119,7 +119,7 @@ namespace NetTopologySuite.Geometries.Implementation
         {
             if (coordSeq == null)
             {
-                Coordinates = new Coordinate[0];
+                Coordinates = Array.Empty<Coordinate>();
                 return;
             }
 
--- a/src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
+++ b/src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
@@ -375,7 +375,7 @@ namespace NetTopologySuite.Geometries.Implementation
             : base(coords?.Length / dimension ?? 0, dimension, measures)
         {
             if (coords == null)
-                coords = new float[0];
+                coords = Array.Empty<float>();
 
             if (coords.Length % dimension != 0)
                 throw new ArgumentException("Packed array does not contain " +
@@ -395,7 +395,7 @@ namespace NetTopologySuite.Geometries.Implementation
         {
             if (coords == null)
             {
-                _coords = new float[0];
+                _coords = Array.Empty<float>();
                 return;
             }
 
--- a/src/NetTopologySuite/IO/WKTReader.cs
+++ b/src/NetTopologySuite/IO/WKTReader.cs
@@ -865,7 +865,7 @@ private Point ReadPointText(TokenStream tokens, GeometryFactory factory, Ordinat
             string nextToken = GetNextEmptyOrOpener(tokens);
             if (nextToken.Equals(WKTConstants.EMPTY))
             {
-                return factory.CreateMultiPoint(new Point[0]);
+                return factory.CreateMultiPoint(Array.Empty<Point>());
             }
 
             // check for old-style JTS syntax (no parentheses surrounding Point coordinates) and parse it if present
--- a/src/NetTopologySuite/IO/WKTReader.cs
+++ b/src/NetTopologySuite/IO/WKTReader.cs
@@ -316,7 +316,8 @@ namespace NetTopologySuite.IO
                 case Ordinates.XY:
                     res = new Coordinate(x, y); break;
                 case Ordinates.XYZ:
-                    res = new CoordinateZ(x, y, z); ; break;
+                    res = new CoordinateZ(x, y, z);
+                    break;
                 case Ordinates.XYM:
                     res = new CoordinateM(x, y, m); break;
                 case Ordinates.XYZM:
--- a/src/NetTopologySuite/Simplify/VWSimplifier.cs
+++ b/src/NetTopologySuite/Simplify/VWSimplifier.cs
@@ -115,7 +115,7 @@ namespace NetTopologySuite.Simplify
                 var inputPts = coords.ToCoordinateArray();
                 Coordinate[] newPts;
                 if (inputPts.Length == 0)
-                    newPts = new Coordinate[0];
+                    newPts = Array.Empty<Coordinate>();
                 else newPts = VWLineSimplifier.Simplify(inputPts, _distanceTolerance);
                 return Factory.CoordinateSequenceFactory.Create(newPts);
             }
--- a/src/NetTopologySuite/Utilities/HexConverter.cs
+++ b/src/NetTopologySuite/Utilities/HexConverter.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Globalization;
 
 namespace NetTopologySuite.Utilities
 {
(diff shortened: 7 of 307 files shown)


hahschaa:
By the way, I was kind of confused by the code in e.g. [BoundaryChainNoder.cs](https://github.com/NetTopologySuite/NetTopologySuite/pull/793/commits/06b665f97cc42c61995c77f408df3f7525ec04e1#diff-5a6d4057b4ebcf5dd578f9673e9cee41a25e5529a9dd4c087dc28dcaad12f13e), the method is called `AddSegments`, but if a segment already exists it is removed, and not re-added afterwards. Is that correct?

hahschaa:
Sorry for the changes after opening the PR. I forgot that this project is also build for .net standard 2.0, so I had to revert some changes that are available in that target framework. But then I some more thing to clean.

airbreather:
> By the way, I was kind of confused by the code in e.g. [BoundaryChainNoder.cs](https://github.com/NetTopologySuite/NetTopologySuite/pull/793/commits/06b665f97cc42c61995c77f408df3f7525ec04e1#diff-5a6d4057b4ebcf5dd578f9673e9cee41a25e5529a9dd4c087dc28dcaad12f13e), the method is called `AddSegments`, but if a segment already exists it is removed, and not re-added afterwards. Is that correct?

It feels like a correct **performance optimization** to me, and your change makes it **even better**.

Per the description in the summary, this is used to find "Boundary segments [...] which are not duplicated in the input polygonal coverage". The "input polygonal coverage" constraint means that any given segment will appear either once or twice in the entire coverage, never more.

It's also correct to use the return value of `HashSet<T>` methods to avoid a redundant `Contains` call (edit: I say "redundant" because both `Remove` and `Add` imply all the same work done by `Contains` — this is a generally applicable optimization that some tools will perform automatically). I don't expect it to be a HUGE win in practice, but it should never make anything worse, so I'm happy with it.

test/NetTopologySuite.Tests.NUnit.Performance/Performance/Operation/Buffer/DepthSegmentStressTest.cs
```diff
@@ -56,11 +56,7 @@ public void RunCompare()
             var seg3 = CreateRandomDepthSegment();
 
             // do test work here
-            bool fails = false;
-            if (!IsSymmetric(seg1, seg2))
-                fails = true;
-            if (!IsTransitive(seg1, seg2, seg3))
-                fails = true;
+            bool fails = !IsSymmetric(seg1, seg2) || !IsTransitive(seg1, seg2, seg3);
```
airbreather: This is a behavior change. `IsTransitive` has side effects that will no longer run if `IsSymmetric` returns `false`, whereas they currently otherwise would run.

A fancy way to resolve this is to change `||` to `|`. A less fancy way is to leave it alone (edit to add:), i.e., revert this change.

src/NetTopologySuite/Utilities/HexConverter.cs
```diff
@@ -22,78 +23,78 @@ private HexConverter() { }
...
-                  // generazione del risultato final
-                  // se il risultato da generare è in base 10 non c'è
-                  // bisogno di calcoli
-                  if (baseOut == 10)
-                           result = valore.ToString();
+                // generazione del risultato final
+                // se il risultato da generare è in base 10 non c'è
+                // bisogno di calcoli
+                if (baseOut == 10)
+                    result = valore.ToString(CultureInfo.InvariantCulture);
```
airbreather, line 79: :thinking: I feel like this probably makes it worse.

This method also contains a `.ToUpper()` and a `double.Parse` which are culture-sensitive, and its two callers also both use unqualified `.ToString()`. Each of these uses the current culture implicitly.

I'm not sure it makes sense to change **only** this one. I would personally **at least** leave such a change for a more focused PR.

*The similar changes in 5234798 do not worry me today because they do not touch the `NetTopologySuite` package.*

src/NetTopologySuite/Algorithm/MinimumBoundingCircle.cs
```diff
@@ -267,7 +267,7 @@ private void ComputeCirclePoints()
             // handle degenerate or trivial cases
             if (_input.IsEmpty)
             {
-                _extremalPts = new Coordinate[0];
+                _extremalPts = Array.Empty<Coordinate>();
```
airbreather, line 270: :page_facing_up: FYI this one in particular is technically a behavior change. I seriously doubt that this will ever matter, and I support the change, but I want to point out these subtleties when I see them:

Because `GetExtremalPoints()` returns this array directly, there will now be cases where `a.GetExtremalPoints() == b.GetExtremalPoints()` when they didn't before, even if `a` and `b` are different.

src/NetTopologySuite/Densify/Densifier.cs
```diff
@@ -159,7 +159,7 @@ protected override CoordinateSequence TransformCoordinates(
                 // prevent creation of invalid LineStrings
                 if (parent is LineString && newPts.Length == 1)
                 {
-                    newPts = new Coordinate[0];
+                    newPts = Array.Empty<Coordinate>();
```
airbreather, line 162: :page_facing_up: Same basic note about behavior change as in `MinimumBoundingCircle.cs`

src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequence.cs
```diff
@@ -64,7 +64,7 @@ public CoordinateArraySequence(Coordinate[] coordinates, int dimension, int meas
         {
             if (coordinates == null)
             {
-                Coordinates = new Coordinate[0];
+                Coordinates = Array.Empty<Coordinate>();
```
airbreather, line 67: :page_facing_up: Same basic note about behavior change as in `MinimumBoundingCircle.cs`

src/NetTopologySuite/Geometries/Implementation/CoordinateArraySequence.cs
```diff
@@ -119,7 +119,7 @@ public CoordinateArraySequence(CoordinateSequence coordSeq)
         {
             if (coordSeq == null)
             {
-                Coordinates = new Coordinate[0];
+                Coordinates = Array.Empty<Coordinate>();
```
airbreather, line 122: :page_facing_up: Same basic note about behavior change as in `MinimumBoundingCircle.cs`

src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
```diff
@@ -375,7 +375,7 @@ public PackedFloatCoordinateSequence(float[] coords, int dimension, int measures
             : base(coords?.Length / dimension ?? 0, dimension, measures)
         {
             if (coords == null)
-                coords = new float[0];
+                coords = Array.Empty<float>();
```
airbreather, line 378: :page_facing_up: Same basic note about behavior change as in `MinimumBoundingCircle.cs`

src/NetTopologySuite/Geometries/Implementation/PackedCoordinateSequence.cs
```diff
@@ -395,7 +395,7 @@ public PackedFloatCoordinateSequence(double[] coords, int dimension, int measure
         {
             if (coords == null)
             {
-                _coords = new float[0];
+                _coords = Array.Empty<float>();
```
airbreather, line 398: :page_facing_up: Same basic note about behavior change as in `MinimumBoundingCircle.cs`

src/NetTopologySuite/IO/WKTReader.cs
```diff
@@ -865,7 +865,7 @@ private MultiPoint ReadMultiPointText(TokenStream tokens, GeometryFactory factor
             string nextToken = GetNextEmptyOrOpener(tokens);
             if (nextToken.Equals(WKTConstants.EMPTY))
             {
-                return factory.CreateMultiPoint(new Point[0]);
+                return factory.CreateMultiPoint(Array.Empty<Point>());
```
airbreather, line 868: :page_facing_up: Same basic note about behavior change as in `MinimumBoundingCircle.cs`

src/NetTopologySuite/Simplify/VWSimplifier.cs
```diff
@@ -115,7 +115,7 @@ protected override CoordinateSequence TransformCoordinates(CoordinateSequence co
                 var inputPts = coords.ToCoordinateArray();
                 Coordinate[] newPts;
                 if (inputPts.Length == 0)
-                    newPts = new Coordinate[0];
+                    newPts = Array.Empty<Coordinate>();
```
airbreather, line 118: :page_facing_up: Same basic note about behavior change as in `MinimumBoundingCircle.cs`

airbreather (changes requested):

test/NetTopologySuite.Tests.NUnit.Performance/Performance/Operation/Buffer/DepthSegmentStressTest.cs
hahschaa: Which side effects do you mean? I only see some compares and a Console.WriteLine. As far as I can see the CompareTo only does integer comparisons without side effects. Technically the console output is a side effect, but is that really important if you already know that IsSymmetric has failed. Furthermore this is in a test, so it should have no influence on any actual operations.

src/NetTopologySuite/Algorithm/MinimumBoundingCircle.cs
hahschaa: This is certainly true, but when comparing references types like arrays, it is important to always consider if you want reference equality versus value equality. My feeling is that when you are comparing a collection of Coordinates, you always want to do some value equality, and only want to check for reference equality in some edge cases.

test/NetTopologySuite.Tests.NUnit.Performance/Performance/Operation/Buffer/DepthSegmentStressTest.cs
airbreather: > Technically the console output is a side effect, but is that really important if you already know that IsSymmetric has failed.

:shrug:

I'm not the one who wrote the `Console.WriteLine` call. I only assume that somebody added it for a reason. I do not know what the reason was, and I am not in a position to spend my time guessing.

[Without knowing that reason, I necessarily cannot claim that the behavior change is appropriate.](https://fs.blog/chestertons-fence)

> Furthermore this is in a test, so it should have no influence on any actual operations.

I have very little doubt that you are correct that this would never matter. I'm also not the one who went out of my way to propose a change to it...

src/NetTopologySuite/Utilities/HexConverter.cs
hahschaa: Whoops, I indeed missed the `ToUpper` and `Parse`. But how does it make things worse?

If I also change those to `.ToUpperInvariant()` and `double.Parse(..., CultureInfo.InvariantCulture);` then the whole thing is not dependent on the locale anymore. I think that is a good thing to have in a non-UI library?

airbreather: To put my comment into context: I'm reviewing a PR entitled "Minor code cleanup", which touches 310 files in ways that are mostly harmless, mostly welcome, and mostly beneficial.

In such a PR, I expected to see **NO** behavior changes; if there **are** any behavior changes, then I was expecting them all to be of the `new T[0]` --> `Array.Empty<T>()` sort, which — while I believe it would not technically be correct for the .NET JIT to apply the same optimization — I'm not going to complain about, other than to call attention to them.

Bringing it back to my specific comment here, now.

---

> Whoops, I indeed missed the `ToUpper` and `Parse`.

And `ToString()` in both callers

> But how does it make things worse?

I don't know of any **specific, real-world** situations where the value of `CurrentCulture` would cause this to behave incorrectly (where it previously behaved correctly). That is why I said " :thinking: I feel like this probably makes it worse" instead of " :warning: This makes it worse".

Mixing `CultureInfo` throughout a single method (and its callers) needs to be justified. Continuing to use a consistent `CultureInfo` throughout a method that has stood as-is for roughly 20 years (or more?), on the other hand, does not IMO need to be justified (though if you **do** think that this should change, then it probably would not be hard to convince me to agree with you, and I would appreciate a **separate** issue and PR with more details about this situation).

> If I also change those to `.ToUpperInvariant()` and `double.Parse(..., CultureInfo.InvariantCulture);` then the whole thing is not dependent on the locale anymore. I think that is a good thing to have in a non-UI library?

(And `ToString()` in both callers, especially if you change the code paths that they would actually hit)

Perhaps. I might even agree with you, and I might have even proposed consistently using `CultureInfo.InvariantCulture` (or a caller-supplied `IFormatProvider`, as is consistent for framework APIs) if I had been reviewing the initial implementation of this before the initial release (knowing what I know today).

But why bury a locale-dependent behavior change in a PR of this nature? This is a public API that has stood for decades, and I personally believe that any behavior changes in such APIs should at least be given attention independently of many other **actually** minor changes.

test/NetTopologySuite.Tests.NUnit.Performance/Performance/Operation/Buffer/DepthSegmentStressTest.cs
hahschaa: Okay, I just was wondering which side effects you mean, because I could not find any. I was just asking if I missed anything.

src/NetTopologySuite/Utilities/HexConverter.cs
hahschaa: Okay, thanks for the clarification. You are indeed correct that this is not the appropriate PR for it.
From my experience at work I know that things like this can lead to very weird bugs when communicating between systems with different locales. So that's why I changed it.

But like I said, you are correct and I will revert it.

test/NetTopologySuite.Tests.NUnit.Performance/Performance/Operation/Buffer/DepthSegmentStressTest.cs
hahschaa: And it is not that I went out of my way to change this mindlessly. As far as I could see there are no side effects. That's why I made the change.

src/NetTopologySuite/Utilities/HexConverter.cs
hahschaa: Furthermore, yes I touches a lot of files, I'm aware. I just wanted to help do some maintenance.
I think even more could be done: There are indentation issues, style issues (e.g. some files prefix fields with an underscore, some fields don't, some fields use `m_`), etc. I can't help you much with the spatial algorithms as that is not my strong suit, but since I checked the code out anyway for trying to come up with a decent solution for [issues 790](https://github.com/NetTopologySuite/NetTopologySuite/issues/790) I thought I could help you with the simple stuff first.

hahschaa:
I think the fence argument is a bit weird and over the top; with the same logic you could say that all the unused using statements were there for a reason and I am to dumb to understand why they are there. But the reality is just that someone forgot to clean up after removing code.

Besides, I did not make these changes willy-nilly, I actually checked if there were any side effects.

But I reverted the changes anyway.

airbreather (approved):

airbreather:
The `license/cla` check is still pending.

@FObermaier as of right now, all of the changes are mechanical and/or require little-to-no actual thought. Should we insist on the CLA, or do you think it's OK to just merge? I personally think it's OK to just merge.

**IMO, don't complete with "Squash and merge"**. Each commit fixes a different thing, so it's IMO more useful for source control history to keep them separated. "Create a merge commit" is my preference here.

hahschaa:
I have agreed to the CLA multiple times now. I have never done this before, so I might be doing something wrong here. I just clicked the button and then get

> You have agreed to the CLA for NetTopologySuite/NetTopologySuite
> 
> Version: 2018-10-10

Is is invalid after I push again?

airbreather:
> I have agreed to the CLA multiple times now. I have never done this before, so I might be doing something wrong here. I just clicked the button and then get
> 
> > You have agreed to the CLA for NetTopologySuite/NetTopologySuite
> > Version: 2018-10-10
> 
> Is is invalid after I push again?

I see — in [the original comment from CLAassistant](https://github.com/NetTopologySuite/NetTopologySuite/pull/793#issuecomment-3069677013), it says this:

> **Harm van Schaaijk** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

That explains why the bot isn't detecting it. I do see your name on the list, so, merging.

Thanks for your contribution

Issue #806, opened 2026-01-13
Migrate to slnx files


bjornharrtell:
Last year Microsoft finished a new solution file format slnx which is a more simple convention based xml format compared to the old visual studio originating sln text file which is hard to read/parse/change manually.

Read more here https://devblogs.microsoft.com/dotnet/introducing-slnx-support-dotnet-cli/.

The process to migrate is simple and should be safe.

PR #809, opened 2026-01-17
Migrate to slnx

bjornharrtell:
ref https://github.com/NetTopologySuite/NetTopologySuite/issues/806


bjornharrtell:
@FObermaier AFAIK this should be a safe change to do at this point. Do you agree?

FObermaier (approved):
~Oftmals~ Looks good to me

PR #829, opened 2026-03-13
Fix ConvexHull.Reduce to avoid modifying input coordinates

### LINKED ISSUE #824: Fix ConvexHull to avoid modifying input geometry (JTS #1083)
## Summary
Fix `ConvexHull` to not mutate the caller's input data, matching JTS 1.21 (#1083).

## Background
JTS fixed `ConvexHull` in [locationtech/jts#1083](https://github.com/locationtech/jts/issues/1083). The algorithm was sorting or otherwise modifying a coordinate array derived from the input geometry. Since `Coordinate[]` arrays in JTS/NTS are mutable references, this could silently modify the caller's geometry coordinates, which is unexpected and incorrect behaviour.

## JTS Source
[`algorithm/ConvexHull.java`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/algorithm/ConvexHull.java)

## NTS File to Update
`src/NetTopologySuite/Algorithm/ConvexHull.cs`

bjornharrtell:
When `ComputeInnerOctolateralRing` returns `null`, `Reduce()` was returning the original input array directly. The caller then sorts this array in-place (via `Array.Sort`), which mutates the caller's input coordinates silently.

Fix: return a clone of the array so the input is never modified.

Ported from JTS commit `73263456b` (Fix ConvexHull to use input clone [JTS #1083](https://github.com/locationtech/jts/issues/1083)).

Closes #824


AsgerPetersen:
I think we should port the test also.

AsgerPetersen (changes requested):
Tests should be ported also

bjornharrtell:
Test case `TestInputUnmodified` added — verifies the input geometry is not modified by ConvexHull computation (a large nearly-linear linestring that triggered the bug). Branch force-pushed.

AsgerPetersen (approved):
Seems to be a correct port. The supplied test fails without the fix and passes with it.

PR #830, opened 2026-03-13
Fix TaggedLineString.GetComponentPoint to prefer result segment vertex

### LINKED ISSUE #825: Fix TopologyPreservingSimplifier: prevent jumping components (JTS #1096)
## Summary
Fix `TopologyPreservingSimplifier` to prevent simplified segments from jumping to incorrect topological components, matching JTS 1.21 (#1096).

## Background
JTS fixed `TopologyPreservingSimplifier` in [locationtech/jts#1096](https://github.com/locationtech/jts/issues/1096). In certain inputs, simplified line segments could be placed in the wrong component of the simplified geometry, resulting in an output with incorrect topology (e.g., lines connected to the wrong ring, or crossing ring boundaries).

## JTS Source
[`simplify/TopologyPreservingSimplifier.java`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/simplify/TopologyPreservingSimplifier.java)

## NTS File to Update
`src/NetTopologySuite/Simplify/TopologyPreservingSimplifier.cs` and related `TaggedLineStringSimplifier`.

bjornharrtell:
`GetComponentPoint()` was always returning `ParentCoordinates[1]` (the second vertex of the unmodified parent line), even after simplification had produced result segments. In `TopologyPreservingSimplifier`, the intersections are computed using this point; if the original vertex was removed by simplification the wrong point was used, which could produce incorrect results.

Fix: when result segments are available, return `_resultSegs[0].P0` (the start vertex of the first simplified segment). Fall back to the original behaviour only when no result segments exist yet.

Ported from JTS commit `917e1da98` (Fix TPS GetComponentPoint [JTS #1096](https://github.com/locationtech/jts/issues/1096)).

Closes #825


AsgerPetersen (changes requested):
Tests should be ported

AsgerPetersen (approved):
I added missing docs and comments and also ported a test that was added later in JTS specifically to test this fix

PR #831, opened 2026-03-13
Fix VWLineSimplifier to copy coordinates in output

### LINKED ISSUE #826: Fix VWSimplifier coordinate aliasing (JTS #1107)
## Summary
Fix a coordinate aliasing bug in `VWSimplifier` (Visvalingam-Whyatt simplifier), matching JTS 1.21 (#1107).

## Background
JTS fixed a coordinate aliasing issue in the Visvalingam-Whyatt simplifier (`VWSimplifier`) in [locationtech/jts#1107](https://github.com/locationtech/jts/issues/1107). The simplifier was reusing coordinate object references rather than copying them, which could cause coordinates in the output geometry to share identity with internal working data. Subsequent mutation of internal state could silently corrupt output geometry coordinates.

## JTS Source
[`simplify/VWSimplifier.java`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/simplify/VWSimplifier.java)

## NTS File to Update
`src/NetTopologySuite/Simplify/VWSimplifier.cs` (and/or internal `VWLineSimplifier.cs`)

bjornharrtell:
`VWLineSimplifier.Simplify()` was inserting direct references to coordinates from the input `_pts` array into the output `CoordinateList`. The returned array therefore aliased the input: any caller that mutated the returned coordinates would silently corrupt the original input array.

Fix: call `.Copy()` on each coordinate when adding it to the result list, so the output array is fully independent of the input.

Ported from JTS commit `d22e30088` (Fix VW Simplifier aliasing [JTS #1107](https://github.com/locationtech/jts/issues/1107)).

Closes #826


AsgerPetersen (changes requested):
Tests should be ported

bjornharrtell:
@AsgerPetersen test is ported now, will you approve?

AsgerPetersen (approved):
Nice. Thank you

PR #832, opened 2026-03-13
Fix buffer erosion artifacts for rings and eroded elements

### LINKED ISSUE #822: Fix BufferOp robustness: hole erosion heuristic and element erosion for negative distance (JTS #1117, #1119)
## Summary
Apply two robustness fixes to `BufferOp` from JTS 1.21: a hole erosion heuristic for ring inputs (#1117) and a fix for element erosion with negative buffer distances (#1119).

### Fix 1: Hole erosion heuristic for rings (#1117)
JTS [locationtech/jts#1117](https://github.com/locationtech/jts/issues/1117) — When buffering ring inputs with a negative distance, buffer holes could be incorrectly eroded, producing unexpected artifacts. A heuristic check was added to detect and remove these artifacts.

### Fix 2: Element erosion for negative distance (#1119)
JTS [locationtech/jts#1119](https://github.com/locationtech/jts/issues/1119) — Negative buffer operations on some geometries produced erroneous non-empty results when the buffer distance was large enough that the geometry should be completely eroded. 

Both fixes improve robustness for negative-distance (inward) buffering.

## NTS Files to Update
`src/NetTopologySuite/Operation/Buffer/BufferOp.cs` and related buffer internals.

bjornharrtell:
Ports two related JTS fixes for buffer robustness issues.

## JTS #1117 — Improved ring erosion heuristic in `BufferCurveSetBuilder`

**Ported from**: JTS commit `040802a74`

Replaces `IsErodedCompletely` with an improved `IsRingFullyEroded` that:

- Takes an `isHole` parameter so the erosion condition is only applied when geometrically meaningful (holes eroded by positive distance, shells eroded by negative distance).
- Degenerate rings (<4 points) are now always considered fully eroded.
- Rewrites `AddRingBothSides` to conditionally skip generating the hole side of a ring buffer when it would be fully eroded, preventing spurious hole artifacts (e.g. [libgeos#1223](https://github.com/libgeos/geos/issues/1223)).

## JTS #1119 — Buffer-specific `PolygonBuilder` removes overlay dependency

**Ported from**: JTS commit `49a6349c3`

- Adds `BufferNodeFactory` in the buffer package (replaces `OverlayNodeFactory` in `BufferBuilder`).
- Adds `Buffer.PolygonBuilder` — functionally identical to `Overlay.PolygonBuilder` except that `PlaceFreeHoles` silently discards holes that do not lie within any shell (these are eroded elements), rather than throwing a `TopologyException`. This fixes buffer failures when a negative-distance buffer erodes a polygon element completely.
- `BufferBuilder` now uses both new buffer-specific classes.

Closes #822


AsgerPetersen (changes requested):
Please also port the jts tests for this

bjornharrtell:
@AsgerPetersen I believe tests are ported now too. I want this in before continuing on https://github.com/NetTopologySuite/NetTopologySuite/pull/838, do you think it's ready for merge?

AsgerPetersen (approved):
Great.

The two remaining minor differences from JTS (AddPolygon coordinate-cleaning order and the missing post-clean empty guard) are pre-existing in NTS and were not introduced by this PR, so they are out of scope here.

PR #834, opened 2026-03-13
Fix OffsetCurve to support mitre joins for polygons

### LINKED ISSUE #815: OffsetCurve: miter joins for polygonal input from JTS 1.21 (JTS #1109)
## Summary
Improve `OffsetCurve` to support miter joins when the input is a polygonal geometry, matching JTS 1.21 (#1109).

## Background
JTS improved `OffsetCurve` in [locationtech/jts#1109](https://github.com/locationtech/jts/issues/1109) to properly handle miter join style when the input is a polygon or multipolygon. Previously the behavior for polygonal inputs with miter joins was inconsistent or fell back to the buffer approach without honoring the join style.

## JTS Source
[`operation/buffer/OffsetCurve.java`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/operation/buffer/OffsetCurve.java)

## NTS File to Update
`src/NetTopologySuite/Operation/Buffer/OffsetCurve.cs`

bjornharrtell:
Port of JTS commit e83b7e40a (Fix OffsetCurve to support mitre joins for polygons #1109).

`OffsetCurveMapOp.Map()` was calling `geom.Buffer(distance)` ignoring the parent's `_bufferParams` when the geometry was a Polygon. This prevented join style settings (e.g. Mitre) from taking effect for polygonal input.

Fix: pass `_bufferParams` through the new `ComputePolygonCurve()` static helper. Also converts `ToLineString` to a static method on the inner class.

Adds `TestPolygonMitreJoin` test case.

Closes #815

### DIFF
--- a/src/NetTopologySuite/Operation/Buffer/OffsetCurve.cs
+++ b/src/NetTopologySuite/Operation/Buffer/OffsetCurve.cs
@@ -193,34 +193,34 @@ namespace NetTopologySuite.Operation.Buffer
                 if (geom is Point) return null;
                 if (geom is Polygon poly)
                 {
-                    return ComputePolygonCurve(poly, _parent._distance, _parent._bufferParams);
+                    return _parent.ComputePolygonCurve(poly, _parent._distance);
                 }
                 return _parent.ComputeCurve((LineString)geom, _parent._distance);
             }
+        }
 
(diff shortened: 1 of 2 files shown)


bjornharrtell:
@AsgerPetersen fixed some cosmetic alignment wrt. upstream and rebased this, so should be ready for review again. :)

PR #835, opened 2026-03-13
Add GeometryCollection dimension cache for performance

### LINKED ISSUE #818: Cache GeometryCollection.Dimension to avoid repeated computation (JTS #1103)
## Summary
Cache the `Dimension` property in `GeometryCollection` to avoid iterating all elements on every call, matching the JTS 1.21 performance improvement (#1103).

## Background
JTS added a dimension cache to `GeometryCollection` in [locationtech/jts#1103](https://github.com/locationtech/jts/issues/1103). Currently in NTS, `GeometryCollection.Dimension` iterates all sub-geometries on every call to compute the maximum dimension. In performance-sensitive code that accesses dimension repeatedly (e.g., RelateNG), this causes unnecessary work.

The fix is to compute the dimension lazily on first call and store it in a backing field.

## NTS File to Update
`src/NetTopologySuite/Geometries/GeometryCollection.cs`

bjornharrtell:
Port of JTS commit 47b52bd02 (Add GeometryCollection dimension cache #1103).

`GeometryCollection.Dimension` and `HasDimension()` were recalculating on every call by iterating all sub-geometries. For large heterogeneous collections this is unnecessarily expensive.

Adds `_hasP`, `_hasL`, `_hasA` boolean fields and `_cachedDimension` field computed lazily on first access via `InitDimCache()`, which uses `GeometryCollectionEnumerator` to traverse nested collections.

Closes #818


bjornharrtell:
Test cases ported. This is trivial enough that I'm confident it can be merged now.

PR #836, opened 2026-03-13
Fix RelateNG to cache in prepared A-L cases

### LINKED ISSUE #819: RelateNG: cache for prepared A-L (area-line) cases (JTS #1099)
## Summary
Add caching in `RelateNG` for the prepared geometry area-line (A-L) case to improve performance, matching JTS 1.21 (#1099).

## Background
JTS added result caching for area-line cases in the prepared geometry `RelateNG` path in [locationtech/jts#1099](https://github.com/locationtech/jts/issues/1099). When a prepared geometry (polygon/area) is tested against many line geometries, reusing cached internal state gives a significant performance boost.

NTS has the RelateNG implementation in `src/NetTopologySuite/Operation/RelateNG/`. The caching improvement should be applied there.

## JTS Source
[`operation/relateng/RelateNG.java`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/operation/relateng/RelateNG.java)

## NTS Files to Update
- `src/NetTopologySuite/Operation/RelateNG/RelateNG.cs`
- Possibly `RelateGeometry.cs` or related classes

bjornharrtell:
Port of JTS commit https://github.com/locationtech/jts/commit/c3d56e6bc66ebd9ae2354600e09687bb54cc1921 (Fix RelateNG to cache in prepared A-L cases #1099).

RelateNG was requiring self-noding (which prevents noder caching) more broadly than needed. This caused poor performance for prepared relate operations involving mixed Area-Line geometry collections.

Changes:
- `RelateGeometry`: add `HasAreaAndLine` property (true when geometry has both areas and lines)
- `RelateGeometry.IsSelfNodingRequired`: add early return when geometry has no areas or lines (e.g. point-only GC)
- `TopologyComputer.IsSelfNodingRequired`: rewrite to check geomB for mixed A/L separately (linear-only B does not require self-noding against polygonal A)

Adds `CheckContainsWithin` and `CheckCoversCoveredBy` assertions to two existing `RelateNGGCTest` tests.

Closes #819

### DIFF
--- a/data/NetTopologySuite.TestRunner.Tests/misc/TestRelateGC.xml
+++ b/data/NetTopologySuite.TestRunner.Tests/misc/TestRelateGC.xml
@@ -577,7 +577,7 @@ GEOMETRYCOLLECTION (POLYGON ((1 9, 5 9, 6 6, 1 5, 1 9), (2 6, 4 8, 6 6, 2 6)), P
 </case>
 
 <case>
-  <desc>GC:AL/A - polygon with overlapping line equal to polygon </desc>
+  <desc>GC:AL/A - polygon with line in boundary and interior equal to polygon </desc>
   <a>
     GEOMETRYCOLLECTION (POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0)), 
       LINESTRING (0 2, 0 5, 5 5))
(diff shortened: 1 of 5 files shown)


bjornharrtell:
@AsgerPetersen added missing test case and docs from upstream and rebased.

Issue #837, opened 2026-03-13
Tests failing / unclear situation


bjornharrtell:
I can see that only NUnit tests are run at CI but that the project contains both Nunit and Xunit tests (NetTopologySuite.Tests.Vivid.XUnit). The NetTopologySuite.Tests.Vivid.XUnit doesn't seem to run without failures:

Test summary: total: 574, failed: 15, succeeded: 251, skipped: 308, duration: 10.6s

bjornharrtell:
@FObermaier / @airbreather do you know why NetTopologySuite.Tests.Vivid.XUnit are not run at CI and/or why they are failing?

bjornharrtell:
@FObermaier I found https://github.com/NetTopologySuite/NetTopologySuite/commit/e6a77693 :)

airbreather:
Missed this ping among the other notifications, sorry

> do you know why NetTopologySuite.Tests.Vivid.XUnit are not run at CI and/or why they are failing?

For my part, one answers the other: by the time I noticed that they're not being run during CI, I think something was already failing, so I didn't bother with it. There's a lot of stuff in this repo that I've personally never needed or wanted, but which (presumably) someone does, so I just put these into that same mental box and didn't think much of it.

bjornharrtell:
IHMO these tests are important to ensure that behaviour of NTS is verified identical to JTS. It's part of the JTS validation suite with lots of very specific/special situations and corner cases.

FObermaier:
The port of the JTS xml-testfile suite is incomplete, that is why some of those tests are failing. Unfortunatly the codebase in NTS is hard to extend.

PR #838, opened 2026-03-13
Enable and fix xunit tests

bjornharrtell:
ref https://github.com/NetTopologySuite/NetTopologySuite/issues/837


bjornharrtell:
https://github.com/NetTopologySuite/NetTopologySuite/issues/822 will probably solve some of the failure cases so I will put this on hold until that is fixed.

PR #849, opened 2026-05-14
[Design record] curves: CircularString, CompoundCurve, CurvePolygon, Triangle, Tin (superseded — see #857)

grootstebozewolf:
**This is no longer the reviewable PR.** The live work is **[#857](https://github.com/NetTopologySuite/NetTopologySuite/pull/857)** — "Add in-tree SQL/MM curve geometry types and WKT I/O foundation", against `develop`. Review that one. The intermediate MVP [#854](https://github.com/NetTopologySuite/NetTopologySuite/pull/854) was closed on 2026-08-04 in favour of #857, which carries the same five types plus SortIndex, ILinearizable, and MultiCurve / MultiSurface.

This PR stays open as the design record that produced it. Three things were settled here: in-tree versus out-of-tree, the `Triangle` naming, and the attribution correction below.

**The dependency on #526 is gone.** This originally sat on top of [#526 "Required Changes for NetTopologySuite.Curve"](https://github.com/NetTopologySuite/NetTopologySuite/pull/526) — @FObermaier's draft, opened 2021-06-10 and last updated 2024-04-11 — for the abstract `Curve` and `Surface<T>` classes and `ILinearizable<T>`. It turned out the prototype needs only a small subset of that foundation, applied to *current* `develop`. That subset passes package validation against the 2.6.0 baseline with a single suppressed diagnostic (CP0012 on `LineString.IsClosed`, virtual → override — binary-compatible in practice). #854 carried that subset and #857 carries it now, foundation commit co-authored with @FObermaier.

**Direction: in-tree.** Per @airbreather's steer — "too much of NTS assumes that all geometry types are implemented fully in-tree" — this work pivoted from *exercising* the abstractions to *porting the out-of-tree prototypes in*. The `CompoundCurve` and `CurvePolygon` prototypes from the [`NetTopologySuite.Curved`](https://github.com/NetTopologySuite/NetTopologySuite.Curved) lineage moved in-tree here, with their structural contract tests. The sub-namespace compromise for `Triangle` stands per the same discussion.

**AI-assisted contribution**, disclosed per [AI_POLICY.md](https://github.com/NetTopologySuite/NetTopologySuite/blob/develop/AI_POLICY.md). Claude assisted with both the code and this description. I have reviewed and tested the result, and I am responsible for its correctness and licensing under the project's usual terms and my signed CLA.

## Attribution: SQL/MM, not OGC SFA

Raised by @airbreather and confirmed against the source. `CIRCULARSTRING`, `COMPOUNDCURVE` and `CURVEPOLYGON` appear in OGC 06-103r4 only as type codes reserved "for future use" — they are not defined types there. They are **SQL/MM Spatial (ISO/IEC 13249-3)** types, and are attributed that way throughout. `Triangle` and `TIN` are the SFA-described types.

## What this adds

Five concrete geometry types in a new `NetTopologySuite.Geometries.Curves` sub-namespace, on the abstract `Curve` and `Surface<T>` classes:

- **`CircularString : Curve`** — SQL/MM curve composed of *n* circular arcs encoded as *2n+1* control points. The constructor accepts an empty coordinate list or an odd count of at least 3, and rejects everything else. Bounds by control points only for now; sagitta-tightening is tracked below.
- **`CompoundCurve : Curve`** *(ported in-tree)* — a single contiguous curve composed of `LineString` / `CircularString` components sharing join points. Flat component list (nested compounds rejected), contiguity enforced in the constructor, chord-based `Length` consistent with `CircularString`.
- **`CurvePolygon : Surface<Curve>`** *(ported in-tree)* — a surface whose ring accessors are typed `Curve` and never collapse to `LinearRing`. Because it extends `Surface<Curve>` rather than `Polygon`, the question of what a legacy `LinearRing`-typed accessor should return never arises here — that is the structural contract the JTS-side epic tracks as F-CP ([locationtech/jts#1195](https://github.com/locationtech/jts/issues/1195), Phase 1 of the SQL/MM Curve Awareness epic). Chord-based `Area`/`Length` fallbacks; `Boundary` returns a `GeometryCollection` of rings until a `MultiCurve` type exists.
- **`Triangle : Surface<LinearRing>`** — OGC SFA Triangle. Constructor takes a 4-coordinate `LinearRing` (3 distinct vertices plus the closing repeat) and rejects other sizes. `SignedDoubleArea` exposes the cross-product signed area. Always has zero interior rings.
- **`Tin : GeometryCollection`** — OGC SFA Triangulated Irregular Network, element type `Triangle`, `Area` the sum of triangle areas. Adjacency and shared-edge invariants are not checked; see open question 4.

Plus `OgcGeometryType.Triangle = 17`; `TIN = 16` was already in the enum. Both match ISO/IEC 13249-3 §5.1.68 Table 15.

**WKT IO.** `WKTReader` and `WKTWriter` natively read and write `CIRCULARSTRING`, `COMPOUNDCURVE`, `CURVEPOLYGON`, `TRIANGLE` and `TIN` through additive dispatch branches; the `ReadOtherGeometryText` / `AppendOtherGeometryTaggedText` extension points remain for external types. SQL/MM tagging rules apply — `LineString` components and rings are written bare, `CircularString` and `CompoundCurve` with their tags — and curves nest inside `GEOMETRYCOLLECTION` through the existing recursive dispatch.

**Build tooling.** The branch also carries the CI migration to the .NET 10 SDK. That is separable from the geometry work and can be split out if a reviewer would rather see it on its own.

**79 new tests** in `test/NetTopologySuite.Tests.NUnit/Geometries/Curves/`:

- `CircularStringTest` (10) — empty / odd / even rejection, NumArcs, IsClosed, endpoints, envelope, copy and reverse round-trip, type-distinct EqualsExact.
- `CompoundCurveTest` (14) — construction validation (null, empty, non-contiguous, nested), join-point sharing, boundary per the Mod-2 rule, envelope union, subtype-preserving copy and reverse.
- `CurvePolygonTest` (19) — the in-tree port of the out-of-tree `CurvePolygonStructuralSpec`, plus construction validation, chord-based area, boundary, and subtype-preserving reverse. The linearization case comes back with the linearization follow-up (open question 3).
- `CurveWktTest` (22) — exact-string WKT round-trip stability for all five types, including EMPTY forms and curves inside `GEOMETRYCOLLECTION`; subtype preservation on read; Z ordinates; rejection of nested `COMPOUNDCURVE`, `TRIANGLE` holes and malformed arcs.
- `TriangleAndTinTest` (14) — 4-coordinate shell enforced, base × height / 2 area, signed-area sign against orientation, collinear degeneracy, copy and reverse, `Tin` holds triangles, area sums.

## Why the namespace split

The existing `NetTopologySuite.Geometries.Triangle` is a long-standing **utility class** (`InCentre`, `IsAcute`, `PerpendicularBisector` on raw `Coordinate`s) that does not participate in the `Geometry` hierarchy. Reclaiming that name would be a breaking rename for every caller across the ecosystem, so the new OGC `Triangle` **geometry** lives in `NetTopologySuite.Geometries.Curves` instead. The two classes serve different roles, and per the discussion the sub-namespace stands as "a pretty good compromise".

## Build status

- Production project (`netstandard2.0`): builds clean on the .NET 10 SDK.
- Test project: full NUnit suite green locally on .NET 10 — 1835 passed, 0 failed.

## Open questions

These carried forward to #857, where several are now resolved.

1. **SortIndex.** All new types currently fall back to the closest existing `SortIndexValue` (LineString / Polygon / MultiPolygon) to avoid mutating `Geometry.cs` on a playground branch. A proper landing should add dedicated entries.
2. **WKB IO.** WKT read and write are in this PR; WKB is not. The out-of-tree repo has `WKBReaderEx` / `WKBWriterEx` prototypes to port next, for extended type codes 8, 9, 10, 16 and 17.
3. **Linearization.** `ILinearizable<T>` is on the foundation, but the in-tree types do not implement it yet. The out-of-tree repo's arc-linearization machinery (`CircularArc`, `ArcSegmentLength`) is the donor. This unblocks the linearization contract test, analytic `Length` and `Area`, and a sagitta-tight `ComputeEnvelopeInternal`. The client-side arc subdivision shipped for SQL Server in NetTopologySuite.IO.SqlServerBytes#35 is a second reference implementation.
4. **`Tin` adjacency invariant.** "The shared edge between any two adjacent triangles is exactly equal as a coordinate pair" is the minimum-viable `IsValid` for a TIN. Not in this PR; it should accompany an eventual landing.
5. **`MultiCurve` / `MultiSurface`.** Present in the out-of-tree repo. Porting them completes the SQL/MM type set and gives `CurvePolygon.Boundary` its proper return type.

The one I would most like reactions on is the porting order for WKB IO against linearization.

## Companion repository

[`NetTopologySuite/NetTopologySuite.Curved`](https://github.com/NetTopologySuite/NetTopologySuite.Curved) — and its active fork [`grootstebozewolf/NetTopologySuite.Curve`](https://github.com/grootstebozewolf/NetTopologySuite.Curve), whose name drops the *d* — served as the out-of-tree incubator. With the in-tree direction settled, that repo is now the **donor** for the remaining ports (IO, linearization, MultiCurve/MultiSurface) rather than a competing long-term home.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -15,16 +15,24 @@ jobs:
     - name: Get source
       uses: actions/checkout@v2
 
-    - name: Setup .NET Core 3.1
+    - name: Setup .NET 10
       uses: actions/setup-dotnet@v5
       with:
-        dotnet-version: 3.1.406
+        dotnet-version: 10
 
     - name: Build
       run: dotnet build -c Release -v minimal -p:WarningLevel=3
 
-    - name: Test
-      run: dotnet test -c Release --no-build --filter '(TestCategory!=LongRunning)&(TestCategory!=Stress)&(TestCategory!=FailureCase)'
+    - name: Test NUnit
+      run: dotnet test test/NetTopologySuite.Tests.NUnit/NetTopologySuite.Tests.NUnit.csproj -c Release --no-build --filter '(TestCategory!=LongRunning)&(TestCategory!=Stress)&(TestCategory!=FailureCase)'
+      shell: bash # defaults disagree on how to quote the filter string
+
+    - name: Test Samples
+      run: dotnet test test/NetTopologySuite.Samples.Console/NetTopologySuite.Samples.Console.csproj -c Release --no-build --filter '(TestCategory!=LongRunning)&(TestCategory!=Stress)&(TestCategory!=FailureCase)'
+      shell: bash # defaults disagree on how to quote the filter string
+
+    - name: Test Vivid XUnit
+      run: dotnet test test/NetTopologySuite.Tests.Vivid.XUnit/NetTopologySuite.Tests.Vivid.XUnit.csproj -c Release --no-build --filter '(TestCategory!=LongRunning)&(TestCategory!=Stress)&(TestCategory!=FailureCase)'
       shell: bash # defaults disagree on how to quote the filter string
 
     - name: Pack
@@ -44,10 +52,10 @@ jobs:
     if: github.event_name == 'push' && (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master')
 
     steps:
-    - name: Setup .NET Core 3.1
+    - name: Setup .NET 10
       uses: actions/setup-dotnet@v5
       with:
-        dotnet-version: 3.1.406
+        dotnet-version: 10
 
     - name: Download Package Files
       uses: actions/download-artifact@v5
@@ -69,10 +77,10 @@ jobs:
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
 
     steps:
-    - name: Setup .NET Core 3.1
+    - name: Setup .NET 10
       uses: actions/setup-dotnet@v5
       with:
-        dotnet-version: 3.1.406
+        dotnet-version: 10
 
     - name: Download Package Files
       uses: actions/download-artifact@v5
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -16,7 +16,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Setup .NET Core 3.1
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v5
       with:
         dotnet-version: 3.1.406
 
@@ -31,7 +31,7 @@ jobs:
       run: dotnet pack -c Release --no-build -o artifacts -p:NoWarn=NU5105
 
     - name: Upload
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v4
       with:
         name: NuGet Package Files (${{ matrix.os }})
         path: artifacts
@@ -45,12 +45,12 @@ jobs:
 
     steps:
     - name: Setup .NET Core 3.1
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v5
       with:
         dotnet-version: 3.1.406
 
     - name: Download Package Files
-      uses: actions/download-artifact@v2
+      uses: actions/download-artifact@v5
       with:
         name: NuGet Package Files (ubuntu-latest)
         path: artifacts
@@ -70,12 +70,12 @@ jobs:
 
     steps:
     - name: Setup .NET Core 3.1
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v5
       with:
         dotnet-version: 3.1.406
 
     - name: Download Package Files
-      uses: actions/download-artifact@v2
+      uses: actions/download-artifact@v5
       with:
         name: NuGet Package Files (ubuntu-latest)
         path: artifacts
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -32,7 +32,7 @@
   <PropertyGroup Condition=" '$(GITHUB_ACTIONS)' == 'True' ">
     <Deterministic>true</Deterministic>
     <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
-    <NtsBuildMetadata Condition=" '$(NtsBuildMetadata)' == '' ">ci.github.$(GITHUB_ACTION)</NtsBuildMetadata>
+    <NtsBuildMetadata Condition=" '$(NtsBuildMetadata)' == '' ">ci.github.$(GITHUB_RUN_ID)</NtsBuildMetadata>
 
     <NtsOfficialRelease Condition=" '$(GITHUB_REF)' == 'refs/heads/master' And '$(GITHUB_EVENT_NAME)' == 'push' ">true</NtsOfficialRelease>
   </PropertyGroup>
--- a/src/NetTopologySuite.TestRunner.Console/NetTopologySuite.TestRunner.Console.csproj
+++ b/src/NetTopologySuite.TestRunner.Console/NetTopologySuite.TestRunner.Console.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net10.0</TargetFramework>
     <RootNamespace>NetTopologySuite</RootNamespace>
     <OutputType>Exe</OutputType>
     <ApplicationIcon>App.ico</ApplicationIcon>
--- a/src/NetTopologySuite/Geometries/Curves/CircularString.cs
+++ b/src/NetTopologySuite/Geometries/Curves/CircularString.cs
@@ -1,0 +1,256 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// AI assistance disclosure:
+//   AI-drafted, human-reviewed and curated. Per the same convention used for the
+//   companion JTS prototype at grootstebozewolf/jts#1, AI-generated portions are
+//   dedicated to CC0-1.0; human curation falls under the NTS BSD-3-Clause grant.
+//
+//   Assisted-by: Claude (Opus-4.7)
+//
+// Status: PLAYGROUND -- prototype of OGC SFA-CA CircularString on top of the
+// foundational Curve/Surface abstractions already on enhancement/curved.
+// Not for merge into develop without further design discussion -- see the PR
+// description for scope, decisions, and open questions.
+
+using System;
+using System.Collections.Generic;
+using NetTopologySuite.Algorithm;
+using NetTopologySuite.Operation;
+
+namespace NetTopologySuite.Geometries.Curves
+{
+    /// <summary>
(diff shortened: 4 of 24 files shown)


airbreather:
I haven't done anything here for quite some time, so my opinions should be weighted accordingly against the opinions of those who have.

> * Whether in-tree (this PR's pattern) or out-of-tree (the `NetTopologySuite.Curved` repo's pattern) is the preferred long-term home.

I would think in-tree is the way to go. IMO too much of NTS assumes that all geometry types are implemented fully in-tree, and so trying to fight that current to reverse the decisions that led there would create probably too much extra work here that would, itself, have a lot of the same drawbacks of putting this in-tree.

The primary drawback of those being how much you still have to deviate from JTS in some fairly core types in order to make it reasonable to have out-of-tree geometry types integrate with the rest of the system.

> * The naming decision in open question #&NoBreak;1.

The usual guiding principle around this is to minimize deviations from JTS in order to maximize the NTS project's ability to adopt future changes that originate from JTS.

It's possible that a project like this will have to deviate from that in many ways, but the sub-namespace seems to be a pretty good compromise, so that's my vote.

> * Whether the AI-assisted draft convention from [grootstebozewolf/jts#1](https://github.com/grootstebozewolf/jts/pull/1) is acceptable on the NTS side, or whether the maintainers would prefer a different disclosure form.

This was the bit that prompted me to respond, which is probably not a surprise.

I see the negative responses that you're getting on locationtech/jts#1195 from very significant contributors to other projects in the broader GIS ecosystem (GDAL and QGIS, respectively). I don't share that sentiment.

However, I would expect accuracy. That issue I linked, for example, claims that "curve geometries — `CIRCULARSTRING`, `COMPOUNDCURVE`, `CURVEPOLYGON`, `MULTICURVE`, `MULTISURFACE`, `TRIANGLE`" are in "OGC SFA / ISO 19125-2". While I do not have direct access to the ISO document(s) in question, I have reviewed OGC 06-103r4 a number of times and skimmed OGC 06-104r4 a bit as well. Revisiting those documents today and doing additional searching, I can find no further publications or revisions to SFA since those were published. So it's disturbing to me to see this list of types despite `Triangle`, `MultiCurve`, and `MultiSurface` being the only three that I can see SFA actually describe.

Especially, a search for the others in the main document shows that the other three names do show up in one place: a table of type code numbers that explicitly mentions that these three "are for future use and do not reflect types used here".

From my searching and scouring, it also seems to me that the ISO standard mentioned in that issue's title was very likely misattributed as well. From https://en.wikipedia.org/wiki/Simple_Features:

> The ISO 19125 standard [...] defines a model for two-dimensional simple features, with linear interpolation between vertices

It then goes on to say:

> The ISO/IEC 13249-3 [SQL/MM Spatial](https://en.wikipedia.org/wiki/SQL/MM_Spatial) extends the Simple Features data model, originally based on [straight-line segments](https://en.wikipedia.org/wiki/Straight-line_segment), adding [circular interpolations](https://en.wikipedia.org/wiki/Circular_interpolation) (e.g. [circular arcs](https://en.wikipedia.org/wiki/Arc_(geometry)))

Even if the contents of the original ISO standard document(s) do describe smooth curves, contrary to my understanding based on what I've been able to find in publicly available sources, I am led to believe that "ISO 19125-2" - the specific part of the standard referred to in the other issue - only addresses SQL bindings for the types described in "ISO 19125-1", so I would not expect it to be linked independently like that.

---

I started looking at this today hoping that I would be able to put in some positive feedback on this emerging work mode and offer any amount of support, but when the title of your strategy document has these (apparent) errors that I caught just by looking at the source that it refers to, I cannot.

airbreather:
> I started looking at this today hoping that I would be able to put in some positive feedback on this emerging work mode and offer any amount of support, but when the title of your strategy document has these (apparent) errors that I caught just by looking at the source that it refers to, I cannot.

Sorry, I should have phrased this a bit differently to say that I **will** happily support this kind of thing if my concerns about accuracy can be addressed. It's not a "no", just a "not yet".

grootstebozewolf:
You are right. I checked OGC 06-103r4 and the table confirms it — CIRCULARSTRING, COMPOUNDCURVE, and CURVEPOLYGON are shaded as future use, not defined types. The attribution should be SQL/MM Spatial (ISO/IEC 13249-3), not ISO 19125-2. The error was in the AI-assisted draft and I missed it. The document is being corrected.

grootstebozewolf:
Pivoted to in-library per the discussion — thanks @airbreather for the clear steer.

**What changed** (commit dc46f0b5, PR description updated to match):

- **`CompoundCurve : Curve`** and **`CurvePolygon : Surface<Curve>`** are now in-tree in `Geometries/Curves`, ported from the out-of-tree `NetTopologySuite.Curve` incubator. `CurvePolygon`'s ring accessors are typed `Curve` and never collapse to `LinearRing` — the F-CP structural contract tests (FCP-S / FCP-MEM / FCP-H / FCP-CP / FCP-DOVE) came along as the regression net.
- The out-of-tree repo is now the **donor** for the remaining ports (WKT/WKB IO, linearization, `MultiCurve`/`MultiSurface`), not a competing home.
- On the accuracy point: the new sources and the PR description attribute the curve types to **SQL/MM Spatial (ISO/IEC 13249-3)** and reserve SFA attribution for `Triangle`/`TIN`, consistent with what OGC 06-103r4 actually defines.

Full NUnit suite on the branch: 1835 passed, 0 failed (.NET 10 SDK).

grootstebozewolf:
Next rung, continuing in-library: **first-class WKT read/write** for the five prototype types (commit 9c16858f, PR description updated).

- `WKTReader`/`WKTWriter` now dispatch `CIRCULARSTRING`, `COMPOUNDCURVE`, `CURVEPOLYGON`, `TRIANGLE` and `TIN` natively, with SQL/MM tagging rules (bare coordinate lists for `LineString` components/rings, tags for the curved ones). The `ReadOtherGeometryText` / `AppendOtherGeometryTaggedText` extension points are untouched for external types.
- This unblocks **FCP-WKT**: the round-trip contract tests from the out-of-tree `CurvePolygonStructuralSpec` now run in-tree — a compound or arc shell keeps its subtype through write → read. Only FCP-TL (linearization) remains deferred.
- 22 new exact-string round-trip cases in `CurveWktTest` (EMPTY forms, Z ordinates, curves inside `GEOMETRYCOLLECTION`, rejection of nested `COMPOUNDCURVE` and `TRIANGLE` holes).

Full NUnit suite: 1859 passed, 0 failed.

grootstebozewolf:
Answering the last open structural question: **the dependency on #526 is gone**. The prototype only needs a small subset of the foundation (Curve, Surface<T>, ILinearizable<T> + the LineString/Polygon base-class rebase applied to *current* develop files), and that subset passes package validation against the 2.6.0 baseline with exactly one suppressed diagnostic (CP0012 on `LineString.IsClosed`, virtual → override — unavoidable and binary-compatible in practice).

The self-contained MVP is now up as **#854** against `develop` (foundation commit co-authored with @FObermaier), verified with the full CI pipeline locally: 2647 + 255 + 255 tests passed, 0 failed, pack green.

Keeping this PR open as the design-conversation record; #854 is the reviewable one.

PR #854, opened 2026-07-20
curves: CircularString, CompoundCurve, CurvePolygon, Triangle, Tin — self-contained MVP with WKT IO

grootstebozewolf:
Self-contained MVP of the curve geometry work, on `develop` — **no longer blocked on #526**. This supersedes the draft #849 as the reviewable PR; #849 stays open as the design-conversation record until this one settles.

> **AI assistance disclosure** *(convention from #849 / [grootstebozewolf/jts#1](https://github.com/grootstebozewolf/jts/pull/1))*: AI-drafted, human-reviewed and curated. AI-generated portions are dedicated to CC0-1.0; human curation falls under the NTS BSD-3-Clause grant.
> `SPDX-License-Identifier: BSD-3-Clause AND CC0-1.0` · `Assisted-by: Claude (Opus-4.7; Fable 5)`

## Why this can stand alone now

The check that motivated this PR: #849 leaned on the dormant #526 branch for the `Curve`/`Surface<T>` foundation. It turns out the prototype only needs a small, well-defined subset of #526, and that subset drops onto current `develop` cleanly:

- **Commit 1** (co-authored with @FObermaier — this is his #526 design, minimally rebased): `Curve`, `Surface<T>`, `ILinearizable<T>` taken as-is; `LineString : Curve` and `Polygon : Surface<LineString>` applied to develop's *current* files (so none of develop's newer work is regressed — the 5-year-old branch's `LineString`/`Polygon` copies are **not** used). Unlike full #526, `EqualsExact`/`CompareTo` semantics are untouched.
- **Commit 2**: the five prototype types (`CircularString`, `CompoundCurve` per SQL/MM Spatial ISO/IEC 13249-3; `CurvePolygon` carrying the F-CP structural contract — ring accessors typed `Curve`, never collapsing to `LinearRing`; OGC SFA `Triangle` and `Tin`) plus first-class WKT read/write with SQL/MM tagging rules.

## Verification (all local runs mirror the CI pipeline exactly)

- `NetTopologySuite.Tests.NUnit`: **2647 passed, 0 failed** (79 curve tests: type behavior, FCP-* contract suite incl. FCP-WKT round-trips, exact-string WKT stability, EMPTY forms, Z ordinates, malformed-input rejection).
- `Samples.Console`: 255 passed, 0 failed. `Tests.Vivid.XUnit`: 255 passed, 0 failed.
[...]

PR #857, opened 2026-08-04
Add in-tree SQL/MM curve geometry types and WKT I/O foundation

grootstebozewolf:
## Summary

Land the SQL/MM curve and surface types as first-class `Geometry` subtypes in the **main** NetTopologySuite library, on new `Curve` / `Surface<T>` bases — not in the retired `NetTopologySuite.Curve` package or submodule.

**Seven geometry types**, all in `NetTopologySuite.Geometries.Curves`:

| Type | Base |
|---|---|
| `CircularString` | `Curve` |
| `CompoundCurve` | `Curve` |
| `CurvePolygon` | `Surface<Curve>` |
| `MultiCurve` | `GeometryCollection` |
| `MultiSurface` | `GeometryCollection` |
| `Triangle` | `Surface<LinearRing>` |
| `Tin` | `GeometryCollection` |

```mermaid
classDiagram
    class Geometry
    class Curve
    class Surface
    class GeometryCollection

    Geometry <|-- Curve
    Geometry <|-- Surface
    Geometry <|-- GeometryCollection

    Curve <|-- LineString
    Curve <|-- CircularString
    Curve <|-- CompoundCurve
    LineString <|-- LinearRing

    Surface <|-- Polygon
    Surface <|-- CurvePolygon
    Surface <|-- Triangle

    GeometryCollection <|-- MultiCurve
    GeometryCollection <|-- MultiSurface
    GeometryCollection <|-- Tin

    class LineString {
      existing - reparented
    }
    class Polygon {
      existing - reparented
    }
    class CircularString
    class CompoundCurve
    class CurvePolygon
    class Triangle
    class MultiCurve
    class MultiSurface
    class Tin
```

**Both I/O paths.** WKT read and write for `CIRCULARSTRING`, `COMPOUNDCURVE`, `CURVEPOLYGON`, `TRIANGLE` and `TIN`, covering EMPTY, multi-component and multi-ring cases, Z ordinates, and subtype-preserving members and rings. WKB read and write for the corresponding SQL/MM type codes, with `WKBGeometryTypes` extended to match ISO/IEC 13249-3 §5.1.68 Table 15.

**`ILinearizable`** provides chord / control-point linearization as an explicit escape hatch. `Linearize()` is the only way to get an approximation; nothing produces one implicitly.

**Fail-closed metrics.** Every metric and analytic member on the curve types — `Length`, `Area`, `Envelope`, `IsSimple`, `Distance`, `Centroid`, `InteriorPoint` — throws `NotSupportedException` naming the type and the member, until the arc-aware follow-up lands. Empty geometries stay exact. This replaces the approximate/placeholder calculations the first revision carried; see the discussion below.

**Dedicated `SortIndexValue` entries** so mixed-type `CompareTo` and `Array.Sort` never reach a wrong `CompareToSameClass` cast.

## Base-class changes — the compatibility-relevant part

`LineString` now extends `Curve`, and `Polygon` extends `Surface<LineString>`. This is the highest-risk change here and the one most worth a careful look: it re-parents two of the most widely used types in the library.

The consequences are captured in `CompatibilitySuppressions.xml` (41 entries), the most notable being `LineString.IsClosed` becoming an `override` via `Curve` — CP0012, binary-compatible in practice. Package validation passes against the 2.6.0 baseline with those suppressions and no others.

## What changed during review

- **Approximate metrics removed.** Per @bjornharrtell: placeholder calculations that could return wrong or misleading results are gone, replaced by the fail-closed `NotSupportedException` behaviour above.
- **Test naming neutralised.** `CurveOracleRedTests` → `CurveMetricsContractTests`, `Curves.Oracle` → `Curves.MetricsContract`, and the assertion and header wording no longer says "Oracle". Tagged `[Category("FailureExpected")]`.

## Relation to existing work

- **[#849](https://github.com/NetTopologySuite/NetTopologySuite/pull/849)** is the design record: it settled in-tree versus out-of-tree, the `Triangle` sub-namespace naming, and the SQL/MM-versus-OGC-SFA attribution. Kept open as history, not for merge.
- **[#854](https://github.com/NetTopologySuite/NetTopologySuite/pull/854)** was the intermediate MVP from `mvp/curves-on-develop`, closed 2026-08-04 in favour of this PR.
- This PR is on current `develop` and adds SortIndex, `ILinearizable`, WKB I/O and the `MultiCurve` / `MultiSurface` types beyond that prototype.
- The [`NetTopologySuite.Curved`](https://github.com/NetTopologySuite/NetTopologySuite.Curved) repository and its fork [`grootstebozewolf/NetTopologySuite.Curve`](https://github.com/grootstebozewolf/NetTopologySuite.Curve) — the name drops the *d* — were the out-of-tree incubator, and remain the donor for follow-up ports rather than a competing home.

## Scope

| In scope | Out of scope |
|---|---|
| Structure, validation, type identity | Arc-aware spatial algorithms |
| WKT read and write | Analytic `Length` / `Area` (these throw) |
| WKB read and write for the SQL/MM type codes | Arc-tight envelopes — bounds are by control point |
| Explicit chord linearization via `Linearize()` | `PolyhedralSurface` (code 15) |
| Dedicated sort indices | Adjacency / shared-edge validity for `Tin` |

## Test plan

- [x] `dotnet build` of the main library and the NUnit test project — 0 errors
- [x] `dotnet test --filter FullyQualifiedName~Geometries.Curves` — structural, WKT, WKB, linearize, CompareTo and fail-closed suites
- [x] Structural: odd control-point count, contiguous `CompoundCurve` members, closed `CurvePolygon` rings
- [x] WKT round-trips with `EqualsExact`, correct keywords, preserved component and ring subtypes
- [x] WKB round-trips cross-checked against GEOS output
- [x] Mixed-type `CompareTo` / `Array.Sort` does not throw
- [x] Fail-closed metrics throw `NotSupportedException` on non-empty curve types; empty stays exact; `Linearize()` remains the only chord path
- [x] No `PackageReference`, `ProjectReference` or `.gitmodules` entry for `NetTopologySuite.Curve`
- [x] Package validation against the 2.6.0 baseline passes with the suppressions listed above

**AI-assisted contribution**, disclosed per [AI_POLICY.md](https://github.com/NetTopologySuite/NetTopologySuite/blob/develop/AI_POLICY.md). Claude assisted with both the code and this description. I have reviewed and tested the result, and I am responsible for its correctness and licensing under the project's usual terms and my signed CLA.

### DIFF
--- a/src/NetTopologySuite/Geometries/Curves/CircularString.cs
+++ b/src/NetTopologySuite/Geometries/Curves/CircularString.cs
@@ -1,0 +1,295 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// AI assistance disclosure:
+//   AI-drafted, human-reviewed and curated. Per the same convention used for the
+//   companion JTS prototype at grootstebozewolf/jts#1, AI-generated portions are
+//   dedicated to CC0-1.0; human curation falls under the NTS BSD-3-Clause grant.
+//
+//   Assisted-by: Claude (Opus-4.7)
+//
+// Status: PLAYGROUND -- prototype of OGC SFA-CA CircularString on top of the
+// foundational Curve/Surface abstractions already on enhancement/curved.
+// Not for merge into develop without further design discussion -- see the PR
+// description for scope, decisions, and open questions.
+
+using System;
+using System.Collections.Generic;
+using NetTopologySuite.Algorithm;
+
+namespace NetTopologySuite.Geometries.Curves
+{
+    /// <summary>
+    /// An OGC SFA-CA <c>CircularString</c>: a sequence of circular arc segments,
+    /// each defined by three consecutive coordinates (start, point on arc, end).
+    /// </summary>
+    /// <remarks>
+    /// A <c>CircularString</c> with <c>2n + 1</c> coordinates encodes <c>n</c> arcs,
+    /// with adjacent arcs sharing endpoints.  An empty <c>CircularString</c> has zero
+    /// coordinates.
+    /// <para/>
+    /// This is a prototype.  In particular, <c>Length</c> and <c>Boundary</c> currently
+    /// fall back to chord-based computations (treating the control points as a polyline)
+    /// rather than analytical arc geometry; the inherited <see cref="Curve.IsClosed"/>
+    /// semantics still apply (start equals end).
+    /// </remarks>
+    [Serializable]
+    public class CircularString : Curve, ILinearizable<LineString>
+    {
+        /// <summary>The control points of the arcs.</summary>
+        private readonly CoordinateSequence _points;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CircularString"/> class.
+        /// </summary>
+        /// <param name="points">The coordinate sequence of control points</param>
+        /// <param name="factory">The geometry factory</param>
+        /// <exception cref="ArgumentException">
+        /// If the sequence has fewer than 3 points or an even number of points
+        /// (must be 0 or odd and at least 3).
+        /// </exception>
+        public CircularString(CoordinateSequence points, GeometryFactory factory) : base(factory)
+        {
+            if (points == null)
+            {
+                points = factory.CoordinateSequenceFactory.Create(0, Ordinates.XY);
+            }
+            if (points.Count != 0)
+            {
+                if (points.Count < 3)
+                {
+                    throw new ArgumentException(
+                        "A non-empty CircularString must have at least 3 control points " +
+                        "(start, on-arc, end of the first arc).", nameof(points));
+                }
+                if (points.Count % 2 == 0)
+                {
+                    throw new ArgumentException(
+                        "A CircularString must have an odd number of control points " +
+                        "(2n + 1 points encode n arcs).", nameof(points));
+                }
+            }
+            _points = points;
+        }
+
+        /// <summary>The control points of this <c>CircularString</c>.</summary>
+        public CoordinateSequence CoordinateSequence => _points;
+
+        /// <summary>The number of arc segments encoded by this <c>CircularString</c>.</summary>
+        public int NumArcs => _points.Count == 0 ? 0 : (_points.Count - 1) / 2;
+
+        /// <inheritdoc cref="Geometry.NumPoints"/>
+        public override int NumPoints => _points.Count;
+
+        /// <inheritdoc cref="Geometry.IsEmpty"/>
+        public override bool IsEmpty => _points.Count == 0;
+
+        /// <inheritdoc cref="Geometry.Coordinate"/>
+        public override Coordinate Coordinate => IsEmpty ? null : _points.GetCoordinate(0);
+
+        /// <inheritdoc cref="Geometry.Coordinates"/>
+        public override Coordinate[] Coordinates => _points.ToCoordinateArray();
+
+        /// <inheritdoc/>
+        public override double[] GetOrdinates(Ordinate ordinate)
+        {
+            if (IsEmpty) return new double[0];
+            var ordinateFlag = (Ordinates)(1 << (int)ordinate);
+            if ((_points.Ordinates & ordinateFlag) != ordinateFlag)
+            {
+                var nulls = new double[_points.Count];
+                for (int i = 0; i < nulls.Length; i++) nulls[i] = Coordinate.NullOrdinate;
+                return nulls;
+            }
+            var vals = new double[_points.Count];
+            for (int i = 0; i < _points.Count; i++) vals[i] = _points.GetOrdinate(i, (int)ordinate);
+            return vals;
+        }
+
+        /// <inheritdoc cref="Curve.StartPoint"/>
+        public override Point StartPoint =>
+            IsEmpty ? null : Factory.CreatePoint(_points.GetCoordinate(0));
+
+        /// <inheritdoc cref="Curve.EndPoint"/>
+        public override Point EndPoint =>
+            IsEmpty ? null : Factory.CreatePoint(_points.GetCoordinate(_points.Count - 1));
+
+        /// <inheritdoc cref="Curve.IsClosed"/>
+        public override bool IsClosed
+        {
+            get
+            {
+                if (IsEmpty) return false;
+                return _points.GetCoordinate(0).Equals2D(_points.GetCoordinate(_points.Count - 1));
+            }
+        }
+
+        /// <inheritdoc cref="Geometry.GeometryType"/>
+        public override string GeometryType => "CircularString";
+
+        /// <inheritdoc cref="Geometry.OgcGeometryType"/>
+        public override OgcGeometryType OgcGeometryType => OgcGeometryType.CircularString;
+
+        /// <summary>
+        /// Length approximated by summing arc-chord lengths.  Analytical arc-length
+        /// computation is tracked in the prototype roadmap.
+        /// </summary>
+        public override double Length => Algorithm.Length.OfLine(_points);
+
+        /// <summary>
+        /// The boundary of a curve per the Mod-2 rule: empty when the curve is empty
+        /// or closed, otherwise the two end points.
+        /// </summary>
+        /// <remarks>
+        /// <c>BoundaryOp</c> only special-cases <see cref="LineString"/> and
+        /// <see cref="MultiLineString"/>; calling it for curve types recurses into
+        /// <see cref="Geometry.Boundary"/> and stack-overflows.  Mirror
+        /// <see cref="CompoundCurve.Boundary"/> until BoundaryOp learns about curves.
+        /// </remarks>
+        public override Geometry Boundary
+        {
+            get
+            {
+                if (IsEmpty || IsClosed)
+                {
+                    return Factory.CreateMultiPoint();
+                }
+                return Factory.CreateMultiPoint(new[] { StartPoint, EndPoint });
+            }
+        }
+
+        /// <inheritdoc/>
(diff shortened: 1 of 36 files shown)


bjornharrtell:
Looks sensible to me. I think the upstream situation is unfortunate...

bjornharrtell:
@grootstebozewolf the upstream issue and non finished PRs are also unfortunate and might be hard to recover from.. far too much information overload compared to this PR. I'm reluctant to introduce this into NTS without upstream support.

grootstebozewolf:
I reached out to @dr-jts have not heard from him.

grootstebozewolf:
Looking at the JTS-family ports for where this foundation can live without waiting on a silent JTS review.

**GEOS already shipped the same Phase-1 bar** (types + WKT/WKB + fail-closed ops) in [libgeos/geos#1046](https://github.com/libgeos/geos/pull/1046) → 3.13, **without** waiting for JTS. Follow-ups are small ops PRs; the only open `Curves` issue today is [libgeos/geos#1497](https://github.com/libgeos/geos/issues/1497) (splitter), not “should we have types?”

I triage-sampled GEOS’s broader backlog (seed `20260808`, 20 of 106 open issues): lots of overlay/buffer/platform debt — **not** a signal to port all of GEOS. Curve **foundation** surface is small; that is the co-upstream for *this* PR’s scope.

If “upstream” means *merged locationtech/jts*, we are still blocked. If it means *a maintained JTS-family structure+WKT implementation*, **GEOS is that**. Happy to keep this PR scoped to that (types + WKT + linearize; no arc overlay).

---
*Co-authored / assisted-by: xAI Grok (grok-4.5). Human reviewed technical claims and scope.*

bjornharrtell:
@grootstebozewolf I was not aware GEOS had implemented something similar, that does change the situation a bit and I think we might consider going ahead.

grootstebozewolf:
### GEOS dovetail: types + WKT/WKB I/O

Landed on this branch (GEOS-aligned, not JTS):

| ISO / GEOS | NTS |
|---|---|
| types 8–10 | `CircularString` / `CompoundCurve` / `CurvePolygon` (existing) + **WKB R/W** |
| types 11–12 | **new** `MultiCurve` / `MultiSurface` + WKT/WKB |
| WKT | `MULTICURVE` / `MULTISURFACE`; tagged `LINESTRING` inside `COMPOUNDCURVE` (GEOS/PostGIS) |

Tests: `FullyQualifiedName~Geometries.Curves` → **119 passed**.

---
*Assisted-by: xAI Grok*

grootstebozewolf:
I'll do one more QA pass, then we are good to go!

grootstebozewolf:
### Oracle RED hooks (signal, not a novel)

Four intentional red tests under `Category=Red` / `Curves.Oracle`:

| Test | Contract |
|---|---|
| Distance centre of unit semicircle | finite, = 1 (`ARC_DISTANCE`) |
| Distance endpoint | 0 |
| Length unit semicircle | π (`ARC_LENGTH`) |
| Envelope past axis | MaxX ≥ 1 on −30°…50° unit arc |

Structural suite still **119 green** (`Category!=Red`). Filter:

```
--filter "Category=Red"          # expect 4 red
--filter "Category!=Red&…Curves" # expect green
```

This is the hook for arc-aware Distance/Length/Envelope so the foundation narrative can stay short.

---
*Assisted-by: xAI Grok*

grootstebozewolf:
@bjornharrtell

I suggest we keep noding / relate / metrics on a separate branch.

This PR is intentionally limited to the foundation (types + structural validation + WKT/WKB + ILinearizable chord path). Once that lands cleanly we can open a follow-up for the operation layer without mixing the two concerns or bloating the review surface here.

Happy to adjust the scope statement if you prefer a different cut.

bjornharrtell:
I agree it make sense scope this to foundation first. I find the language output from grok confusing. What does "Oracle RED hooks (signal, not a novel)" mean and what is this reference to Oracle in the first place?

grootstebozewolf:
@bjornharrtell Good catch — that earlier heading was poorly worded on my side.

**What "Oracle" meant here (not the database):**
In this ecosystem the word is already overloaded (see `NetTopologySuite.IO.Oracle`). The formal reference used for differential testing lives in the companion [NetTopologySuite.Proofs](https://github.com/grootstebozewolf/NetTopologySuite.Proofs) project. The extracted executable the C# side compares against is called **RocqRefRunner**.

**What those four red tests are:**
`Category=Red` / `Curves.Oracle` are only lightweight, intentionally failing placeholders that pin the expected arc-aware contracts for Distance / Length / Envelope. They are a signal for the follow-up metric work so the foundation PR can stay short — not an introduction of RocqRefRunner (or any formal-ref machinery) into core NTS. Structural/WKT coverage remains green under `Category!=Red`.

Happy to reword any remaining "oracle" mentions in the PR description, comments, or category names to "RocqRefRunner" / "formal reference" / something blander if that reduces confusion.

bjornharrtell:
@grootstebozewolf I see, yes I would like to see a more generic/neutral name for these tests.

grootstebozewolf:
@bjornharrtell Renamed the intentional-fail metric hooks to neutral names:

- `CurveOracleRedTests` → `CurveMetricsContractTests`
- `Curves.Oracle` → `Curves.MetricsContract`
- assert / header wording no longer says "Oracle"

Also tagged them `[Category("FailureCase")]` so the existing full-ci filter excludes them (same pattern as other known-fail fixtures). They remain runnable with:

` 
--filter TestCategory=Curves.MetricsContract
` 

Default curves suite under the CI filter is green (119 passed).

src/NetTopologySuite/Geometries/Curves/CircularString.cs
```diff
@@ -0,0 +1,294 @@
...
+                if (IsEmpty || IsClosed)
+                {
+                    return Factory.CreateMultiPoint();
+                }
+                return Factory.CreateMultiPoint(new[] { StartPoint, EndPoint });
+            }
+        }
+
+        /// <inheritdoc/>
+        protected override Envelope ComputeEnvelopeInternal()
```
bjornharrtell, line 165: I should be noted in public callers i.e GetEnvelope that the result is a placeholder and not the true / correct result for curve types.

bjornharrtell:
@grootstebozewolf I'm getting resistance accepting this. One argument against is the "approximate" or "placeholder" calculations, that can give wrong/misleading results, which we're not sure make sense. Perhaps they should simply throw indicating they are not implemented at this point.

grootstebozewolf:
@bjornharrtell Approximate/placeholder calculations are gone.

Every metric/analytic member on the five curve types (`CircularString`, `CompoundCurve`, `CurvePolygon`, `MultiCurve`, `MultiSurface`) now throws `NotSupportedException` with a message that names the follow-up and `Linearize()` as the explicit chord escape hatch. The cut is unconditional — including all-linear content (e.g. a `CompoundCurve` of only `LineString`s). Empty geometries keep their trivial exact values (`Length`/`Area` = 0, empty `Envelope`).

Specifics that address the silent-wrong-answer cases:
- `IsSimpleOp` chord fallback is reverted; `IsSimple` / `IsRing` throw
- `Distance` / `IsWithinDistance` / `NearestPoints` fail closed from either operand (no more `double.MaxValue` sentinel), including curves nested in a plain `GeometryCollection`
- `Centroid` throws instead of NRE; `InteriorPoint` throws instead of silently dropping the curve
- `MultiCurve.Boundary` / `MultiSurface.Boundary` throw `NotSupportedException` instead of `AssertionFailedException`

What still works: structure/validation, WKT/WKB I/O, `Coordinates`/`CoordinateSequence` exposure, `EqualsExact`, `CompareTo`, curve `Boundary` (mod-2 endpoints), `Copy`, `Reverse`, parameterless `Linearize()`, empty-geometry trivial values, and everything on `LineString` / `LinearRing` / `Polygon` (throws are not on the `Curve` / `Surface<T>` bases).

bjornharrtell (dismissed):

grootstebozewolf:
Reset this branch to the foundation cut we agreed on 8–9 Aug: types, structural validation, WKT/WKB, ILinearizable, and fail-closed metrics.

The 30 Aug exact Length / Envelope / Distance and validity/simplicity work is parked on feat/curves-metrics-followup and will come as a follow-up PR after this lands.

CLA is clean again (Claude is no longer a committer on this branch).

@bjornharrtell @AsgerPetersen @FObermaier — is this foundation slice acceptable to merge?

bjornharrtell:
@grootstebozewolf we have multiple concerns about proceeding, amongst others the already mentioned conderation of https://github.com/NetTopologySuite/NetTopologySuite/pull/526, other design considerations, if it warrants a major version and indeed if we currently at all have the needed resources and willingness to adopt and maintain this extension of NTS scope.

grootstebozewolf:
@bjornharrtell this slice is the same Phase-1 bar GEOS already shipped in 3.13 ([libgeos/geos#1046](https://github.com/libgeos/geos/pull/1046)): types, WKT/WKB, fail-closed ops. Dan Baston called it the minimum that could safely go into master — create/read/write for CircularString, CompoundCurve, CurvePolygon, MultiCurve, MultiSurface, and error out on the rest.

That has been in a stable GEOS release since September 2024. Follow-up curve work there has been ops ([#1104](https://github.com/libgeos/geos/pull/1104), [#1106](https://github.com/libgeos/geos/pull/1106), later length/linearize), not a retreat from having the types. Capacity to carry the foundation did not turn out to be the problem.

If the objection is owning later arc-aware metrics in NTS, that is a separate cut and I can keep it off `develop`. If the objection is the foundation itself, that is the bar GEOS already accepted without waiting on JTS.

Issue #853, opened 2026-07-15
Use ReadOnlySpans instead of arrays in methods' arguments.


aAndrzej-dev:
Some methods, such as the LineString constructor, only allow Coordinate[], which is later passed to the CoordinateSequenceFactory.Create method. The method only iterates over the array and does not expose the array elsewhere (except GetCommonSequenceParameters, which can also be simply modified to use ReadOnlySpan). That way, creating a new function (or replacing the existing one as arrays are implicitly converted to ReadOnlySpans) would make creating LineString more efficient when creating it from a slice of a list of coordinates. Moreover it can applief to more public function that do interations or simple operations over arrays.
Functions I suggest to add:
- in CoordinateSequenceFactory:
```cs
public virtual CoordinateSequence Create(ReadOnlySpan<Coordinate> coordinates)
public static (int Count, int Dimension, int Measures) GetCommonSequenceParameters(ReadOnlySpan<Coordinate> coordinates)
```
- in LineString:
```cs
public LineString(ReadOnlySpan<Coordinate> points) : this(DefaultFactory.CoordinateSequenceFactory.Create(points), DefaultFactory)
```

bjornharrtell:
Thanks for the suggestion @aAndrzej-dev, intuitively it seems like something that could be a very significant optimization (depending on use case).

PR #859, opened 2026-08-24
Add ReadOnlySpan<Coordinate> overloads to CoordinateSequenceFactory and GeometryFactory

bjornharrtell:
## Summary

Addresses #853 (request for `ReadOnlySpan<Coordinate>` overloads to avoid unnecessary array allocations), while keeping geometry creation flowing through the owning `GeometryFactory` rather than a hardcoded default factory as originally suggested in the issue.

## Changes

- `CoordinateSequenceFactory`: added `Create(ReadOnlySpan<Coordinate>)` and `GetCommonSequenceParameters(ReadOnlySpan<Coordinate>)`. The existing `Coordinate[]` overloads now forward to the span versions (arrays convert implicitly), removing duplicated logic.
- `GeometryFactory`: added `CreateLineString(ReadOnlySpan<Coordinate>)` and `CreateLinearRing(ReadOnlySpan<Coordinate>)`.
- `CoordinateArraySequenceFactory`, `PackedCoordinateSequenceFactory`, `DotSpatialAffineCoordinateSequenceFactory`: each now overrides the span overload with a specialized implementation instead of falling back to the generic `SetOrdinate`-loop in the base class.
- The internal `PackedDoubleCoordinateSequence`/`PackedFloatCoordinateSequence` constructors now accept `ReadOnlySpan<Coordinate>` instead of `Coordinate[]` (these are `internal`, so this is not a breaking change), avoiding an extra array allocation when a caller already has a span.

All changes are additive/non-breaking. Existing `Coordinate[]`-based call sites are unaffected (arrays convert to spans at zero cost).

## Testing

- `dotnet build` succeeds for both `netstandard2.0` and `netstandard2.1`.
- Full `NetTopologySuite.Tests.NUnit` suite passes (2570 passed, 0 failed, 25 skipped).
- Added `GeometryFactoryTest.TestCreateLineStringFromSpan` verifying parity between the array and span overloads.

## AI disclosure

This PR (code changes, commit message, and this description) was produced with the assistance of an AI coding agent (GitHub Copilot), based on investigation of #853, and reviewed by me before submission.

## Benchmarks

Added a `test/NetTopologySuite.Benchmark` project (BenchmarkDotNet) comparing the new `ReadOnlySpan<Coordinate>` path against the previous "extract a sub-array first" approach. Run with:

```
dotnet run --project test/NetTopologySuite.Benchmark/NetTopologySuite.Benchmark.csproj -c Release
```

Environment: .NET 10.0.11, Ubuntu 26.04, AMD Ryzen 7 PRO 8840U (ShortRun job: 3 iterations/3 warmup, for quick turnaround — not a full statistically-rigorous run).

### Generic base-class loop vs specialized `PackedCoordinateSequenceFactory` override

Confirms the value of giving each built-in factory its own `Create(ReadOnlySpan<Coordinate>)` override instead of relying on the base class's per-ordinate `SetOrdinate` loop:

| Method                     | Count | Mean       | Ratio | Allocated |
|--------------------------- |------ |-----------:|------:|----------:|
| GenericBaseImplementation  | 10    |    98.5 ns |  1.00 |     248 B |
| SpecializedPackedOverride  | 10    |    61.1 ns |  0.62 |     248 B |
| GenericBaseImplementation  | 100   |   899.1 ns |  1.00 |    1688 B |
| SpecializedPackedOverride  | 100   |   599.1 ns |  0.67 |    1688 B |
| GenericBaseImplementation  | 1000  | 9,322.7 ns |  1.00 |   16088 B |
| SpecializedPackedOverride  | 1000  | 5,376.1 ns |  0.58 |   16088 B |

~35-40% faster, same allocations (as expected — this comparison isolates CPU cost, not allocations).

### Extract-then-create vs `Create(ReadOnlySpan<Coordinate>)`, per built-in factory

Slices 10/100/1000 coordinates out of a larger array, either via `CoordinateArrays.Extract` (old approach) or directly via the new span overload:

| Factory                                 | Alloc Ratio (span vs extract) |
|----------------------------------------- |:-----------------------------:|
| `CoordinateArraySequenceFactory`          | 1.00 (no change)              |
| `DotSpatialAffineCoordinateSequenceFactory` | 1.00 (no change)            |
| `PackedCoordinateSequenceFactory`          | ~0.67-0.70 (30-33% less)       |

`CoordinateArraySequenceFactory` and `DotSpatialAffineCoordinateSequenceFactory` still need to materialize an owned `Coordinate[]`/internal array, so slicing doesn't reduce allocations for them. `PackedCoordinateSequenceFactory` benefits directly since it can build its packed `double[]`/`float[]` straight from the span, skipping the intermediate `Coordinate[]` entirely.

### `GeometryFactory.CreateLineString`, default factory (`CoordinateArraySequenceFactory`)

| PointCount | ExtractThenCreateLineString | CreateLineStringFromSpan | Alloc Ratio |
|-----------:|----------------------------:|-------------------------:|:-----------:|
| 10         | 56.1 ns / 208 B              | 55.0 ns / 208 B           | 1.00        |
| 100        | 344.8 ns / 928 B             | 355.4 ns / 928 B          | 1.00        |
| 1000       | 3,445.9 ns / 8128 B          | 4,048.9 ns / 8128 B       | 1.00        |

No allocation win here (same reasoning as above — the default `CoordinateArraySequenceFactory` needs its own array either way) and the span overload is not consistently faster at this call level; the real gains are for `PackedCoordinateSequenceFactory` and for the internal `SetOrdinate`-loop avoidance shown above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ bld/
 # Roslyn cache directories
 *.ide/
 
+# BenchmarkDotNet
+BenchmarkDotNet.Artifacts/
+
 # MSTest test Results
 [Tt]est[Rr]esult*/
 [Bb]uild[Ll]og.*
--- a/NetTopologySuite.slnx
+++ b/NetTopologySuite.slnx
@@ -42,6 +42,7 @@
     <File Path="doc/overwrite/namespaces/NetTopologySuite.Utilities.md" />
   </Folder>
   <Folder Name="/NetTopologySuite.Tests/">
+    <Project Path="test/NetTopologySuite.Benchmark/NetTopologySuite.Benchmark.csproj" />
     <Project Path="test/NetTopologySuite.Samples.Console/NetTopologySuite.Samples.Console.csproj" />
     <Project Path="test/NetTopologySuite.TestRunner/NetTopologySuite.TestRunner.csproj" />
     <Project Path="test/NetTopologySuite.Tests.NUnit.Performance/NetTopologySuite.Tests.NUnit.Performance.csproj" />
--- a/src/NetTopologySuite/Geometries/CoordinateSequenceFactory.cs
+++ b/src/NetTopologySuite/Geometries/CoordinateSequenceFactory.cs
@@ -38,36 +38,44 @@ namespace NetTopologySuite.Geometries
         /// <returns>A coordinate sequence.</returns>
         public virtual CoordinateSequence Create(Coordinate[] coordinates)
         {
+            return Create((ReadOnlySpan<Coordinate>)coordinates);
+        }
+
+        /// <summary>
+        /// Returns a <see cref="CoordinateSequence" /> based on the given span of coordinates;
+        /// whether or not the values are copied is implementation-dependent.
+        /// </summary>
+        /// <param name="coordinates">A span of coordinates, which may not contain null elements</param>
+        /// <returns>A coordinate sequence.</returns>
+        public virtual CoordinateSequence Create(ReadOnlySpan<Coordinate> coordinates)
+        {
             (int count, int dimension, int measures) = GetCommonSequenceParameters(coordinates);
             var result = Create(count, dimension, measures);
-            if (coordinates != null)
+            int spatial = dimension - measures;
+            for (int i = 0; i < coordinates.Length; i++)
             {
-                int spatial = dimension - measures;
-                for (int i = 0; i < coordinates.Length; i++)
+                var coord = coordinates[i];
+                int coordDimension = Coordinates.Dimension(coord);
+                int coordMeasures = Coordinates.Measures(coord);
+                int coordSpatial = coordDimension - coordMeasures;
+                for (int dim = 0; dim < coordSpatial; dim++)
                 {
-                    var coord = coordinates[i];
-                    int coordDimension = Coordinates.Dimension(coord);
-                    int coordMeasures = Coordinates.Measures(coord);
-                    int coordSpatial = coordDimension - coordMeasures;
-                    for (int dim = 0; dim < coordSpatial; dim++)
-                    {
-                        result.SetOrdinate(i, dim, coord[dim]);
-                    }
+                    result.SetOrdinate(i, dim, coord[dim]);
+                }
 
-                    for (int dim = coordSpatial; dim < spatial; dim++)
-                    {
-                        result.SetOrdinate(i, dim, Coordinate.NullOrdinate);
-                    }
+                for (int dim = coordSpatial; dim < spatial; dim++)
+                {
+                    result.SetOrdinate(i, dim, Coordinate.NullOrdinate);
+                }
 
-                    for (int measure = 0; measure < coordMeasures; measure++)
-                    {
-                        result.SetOrdinate(i, spatial + measure, coord[coordSpatial + measure]);
-                    }
+                for (int measure = 0; measure < coordMeasures; measure++)
+                {
+                    result.SetOrdinate(i, spatial + measure, coord[coordSpatial + measure]);
+                }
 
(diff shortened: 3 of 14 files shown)


bjornharrtell:
## Real-world benefit analysis

Looked at this in more depth since the numbers in the PR description are ratios/nanoseconds and easy to misread as "free perf win everywhere." Summary of where the benefit actually lands:

**1. Changes are purely additive — zero benefit until call sites are rewritten.**
Existing `Coordinate[]`-based code is unaffected either way (same allocations, same code path via the array→span implicit conversion), except for one indirect win below. No regressions, but also no automatic gains for the existing codebase or downstream consumers who don't touch their code.

**2. The default factory (`CoordinateArraySequenceFactory`, used by `GeometryFactory.Default`) sees no allocation win from spans.**
`CoordinateArraySequenceFactory.Create(ReadOnlySpan<Coordinate>)` still does `coordinates.ToArray()` internally because `CoordinateArraySequence` is backed by an owned `Coordinate[]`. Same story for `DotSpatialAffineCoordinateSequenceFactory`. The `GeometryFactory.CreateLineString` benchmark table confirms this: allocation ratio is 1.00 at every size, and at 1000 points the span path was actually *slower* (4048.9 ns vs 3445.9 ns for extract-then-create) — likely bounds-check/inlining noise from the extra indirection, but it shows there's no free lunch here. Since `CoordinateArraySequenceFactory` is what most NTS users hit by default, **most consumers get an API convenience (skip writing your own `CoordinateArrays.Extract`/`ToArray` call) but no measurable perf or GC benefit.**

**3. The real win is narrow: `PackedCoordinateSequenceFactory` users.**
Only `PackedCoordinateSequenceFactory` can build its packed `double[]`/`float[]` directly from a span, skipping the intermediate `Coordinate[]` entirely — that's the ~30-33% allocation reduction in the slice benchmark. This benefits:
   - New code that already has a span (e.g. slicing a large buffer while parsing WKB/shapefiles/etc.) and adopts the new overload directly.
   - **Existing** `Coordinate[]`-based callers of `PackedCoordinateSequenceFactory.Create(Coordinate[])` for free, since it now forwards to the specialized span override instead of the generic base-class `SetOrdinate` loop — this is the ~35-40% CPU improvement in the first benchmark table, and it applies without any caller code changes.

**4. Magnitude is small in absolute terms.**
Even in the best case (Packed factory, 1000 coordinates), the CPU delta is on the order of ~4 μs and the allocation delta is on the order of a few/tens of KB per geometry. That's not something a typical desktop/GIS application creating dozens-to-thousands of geometries will notice. It only compounds into something visible in high-volume, GC-pressure-sensitive scenarios: bulk ingestion pipelines (parsing millions of features from shapefile/WKB/WKT), streaming spatial index builds, or similar hot paths — and only if they're using `PackedCoordinateSequenceFactory`.

### Conclusion
- No regression risk; safe, additive API surface addressing #853 as requested.
- Real, measurable benefit is limited to `PackedCoordinateSequenceFactory` consumers: ~35-40% less CPU (automatic, no code change needed) and ~30% less allocation (requires adopting the span overload) per sequence creation.
- For the default `CoordinateArraySequenceFactory` (and `DotSpatialAffineCoordinateSequenceFactory`), the new overloads are ergonomic sugar only — no measurable perf/GC improvement, since those sequences are still backed by an owned array that must be materialized regardless of array or span input.
- Practical impact will only be visible for large-scale/bulk geometry-construction workloads that already use (or switch to) `PackedCoordinateSequenceFactory` with span-based slicing; typical usage of the library (default factory, moderate geometry counts) won't see a measurable difference.

bjornharrtell:
@AsgerPetersen not sure it's worth it, considering analysis and conclusions above... but then again, it's small API addition and could be significant in some specific use cases (nudge nudge?)?

PR #870, opened 2026-08-28
Fix OverlayEdge.AddCoordinates skipping/duplicating first point (JTS #1187)

### LINKED ISSUE #863: OverlayNG: fix OverlayEdge.AddCoordinates skipping first point (JTS #1187)

`OverlayEdge.AddCoordinates` (`src/NetTopologySuite/Operation/OverlayNG/OverlayEdge.cs:105`) has:

```csharp
bool isFirstEdge = coords.Count > 0;
```

This is inverted — it should be `coords.Count == 0` (i.e. `coords.IsEmpty`). JTS had the identical bug and fixed it in [locationtech/jts#1187](https://github.com/locationtech/jts/pull/1187) / [commit 52c5d9882](https://github.com/locationtech/jts/commit/52c5d988237367a01800f90b7f4ebbd9de60f8a9):

```java
-    boolean isFirstEdge = coords.size() > 0;
+    boolean isFirstEdge = coords.isEmpty();
```

With the current (buggy) NTS logic, the first coordinate of a ring's coordinate sequence is skipped only when the output list is *not* empty, and included when it *is* empty — backwards from the intended "include the very first point of the very first edge, skip duplicate first points on subsequent edges" behavior. This can cause a missing start coordinate in `OverlayNG` result assembly for certain edge orderings.

Port the one-line fix and add a regression test matching JTS's added case in `OverlayNGTest`.

bjornharrtell:
Closes #863.

Ports JTS commit [`52c5d9882`](https://github.com/locationtech/jts/commit/52c5d988237367a01800f90b7f4ebbd9de60f8a9) (`OverlayEdge.java`, JTS locationtech/jts#1187).

## Problem
`OverlayEdge.AddCoordinates` computed:
```csharp
bool isFirstEdge = coords.Count > 0;
```
which is the inverse of the intended condition. This skips the very first coordinate of the very first edge added to a result ring/line, and (for subsequent edges) fails to skip the duplicate start point — causing `OverlayNG` result rings to come out rotated to start at the wrong vertex.

## Fix
`isFirstEdge = coords.Count == 0` (i.e. `coords.IsEmpty`), matching the one-line JTS fix.

## Tests
Ported JTS's regression test as `TestRingsNonRotated_JTS1187` in `OverlayNGTest.cs` — intersecting a polygon with itself and checking the result ring is *exactly* equal (same starting vertex), not just topologically equal.

Verified this test **fails without the fix** (`POLYGON ((1 9, 9 9, 9 1, 1 1, 1 9))` → rotated to `POLYGON ((9 9, 9 1, 1 1, 1 9, 9 9))`) and passes with it. Full `OverlayNGTest` suite: 52 passed, 8 pre-existing skips.

---
**AI disclosure:** This fix was ported and this PR prepared with AI assistance (GitHub Copilot CLI), based on a direct diff against JTS commit 52c5d9882. Please review before merging.

### DIFF
--- a/src/NetTopologySuite/Operation/OverlayNG/OverlayEdge.cs
+++ b/src/NetTopologySuite/Operation/OverlayNG/OverlayEdge.cs
@@ -102,7 +102,8 @@ namespace NetTopologySuite.Operation.OverlayNG
         /// <param name="coords">The coordinate list to add to</param>
         public void AddCoordinates(CoordinateList coords)
         {
-            bool isFirstEdge = coords.Count > 0;
+            //-- fixes https://github.com/locationtech/jts/pull/1187 (JTS commit 52c5d9882)
+            bool isFirstEdge = coords.Count == 0;
             if (IsForward)
             {
                 int startIndex = 1;
--- a/test/NetTopologySuite.Tests.NUnit/Coverage/CoverageGapFinderTest.cs
+++ b/test/NetTopologySuite.Tests.NUnit/Coverage/CoverageGapFinderTest.cs
@@ -23,7 +23,32 @@ namespace NetTopologySuite.Tests.NUnit.Coverage
             var coverage = ToArray(covGeom);
             var actual = CoverageGapFinder.FindGaps(coverage, gapWidth);
             var expected = Read(wktExpected);
-            CheckEqual(expected, actual);
+            /*
+             * CoverageGapFinder returns gap rings as (closed) LineStrings.
+             * LineString.Normalized() does not rotate the ring to a canonical
+             * start point/orientation the way Polygon/LinearRing normalization
+             * does, so comparing rings directly as LineStrings is fragile with
+             * respect to which vertex happens to be first.
+             * Wrap them as polygons (test-only) for a robust, rotation- and
+             * orientation-invariant comparison. JTS made the equivalent fix to
+             * its own port of this test (see JTS commit e0bddb56e).
+             */
+            CheckEqual(RingsAsPolygons(expected), RingsAsPolygons(actual));
+        }
+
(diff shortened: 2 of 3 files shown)


bjornharrtell:
CI was initially failing on `CoverageGapFinderTest.TestThreePolygonGap` — a pre-existing test that compared gap rings as plain `LineString`s, which don't get canonical rotation/orientation normalization the way `Polygon`/`LinearRing` do. This fix legitimately changes which vertex starts the affected ring (correctly, matching JTS), which exposed that fragility. Fixed by making the test comparison rotation/orientation-invariant (matching how JTS fixed the equivalent issue in its own test suite, commit `e0bddb56e`), without changing `CoverageGapFinder`'s public API. CI is now green.

fallenmi (approved):
AI disclosure: I used OpenAI Codex to assist this review; I verified the results locally against exact head `41aead34fd24a129f0a84f93d8534553d328e6a0`.

The predicate now matches JTS commit `52c5d9882`, and the CoverageGapFinder test adjustment matches the intent of JTS commit `e0bddb56` while keeping NTS's LineString-returning API unchanged.

I added an independent two-case oracle around `OverlayEdge.AddCoordinates` that exercises both orientations and a joined second edge: exact base `e064b57b` fails both cases because the first endpoint is skipped, while this head passes both. The focused OverlayNG/Coverage tests pass 53 with 8 existing skips; the full NUnit suite passes 2,573 with 25 skips, including the two review-oracle cases. GitHub's Windows, macOS, and Linux builds and the CLA check are green.

Looks good to me.

AsgerPetersen (approved):
PR seems to faithfully port the changes from https://github.com/locationtech/jts/pull/1187

PR #871, opened 2026-08-28
Make HalfEdge.CompareAngularDirection / Quadrant computation more robust (JTS #1224, #1226)

### LINKED ISSUE #864: EdgeGraph: make HalfEdge.CompareAngularDirection / Quadrant computation more robust (JTS #1224, #1226)

NTS `HalfEdge.CompareAngularDirection` (`src/NetTopologySuite/EdgeGraph/HalfEdge.cs:404`) still uses the direction-vector-based algorithm (`DirectionX`/`DirectionY` + `Quadrant.Quadrant(dx, dy)`) that JTS replaced for robustness:

- [locationtech/jts#1224](https://github.com/locationtech/jts/commit/a3927e635) — `HalfEdge.compareAngularDirection` now compares `directionPt()` equality directly and derives the quadrant from `(orig(), directionPt())` endpoints rather than a separately-computed `(dx, dy)` vector, avoiding roundoff error from an intermediate subtraction.
- [locationtech/jts#1226](https://github.com/locationtech/jts/commit/bc1be3d42) — `Quadrant` gains an endpoint-based overload and `PolygonNodeTopology`/`EdgeEnd`/`DirectedEdge` are updated to use it instead of computing raw vector components first.

Port both changes: add the endpoint-based `Quadrant` overload, update `HalfEdge.CompareAngularDirection` to use it, and update any other NTS call sites (`PolygonNodeTopology`, `EdgeEnd`, `DirectedEdge` equivalents) that currently compute `(dx, dy)` manually before calling `Quadrant`.

bjornharrtell:
Closes #864.

Ports JTS commits [`a3927e635`](https://github.com/locationtech/jts/commit/a3927e635dfae12853f423db2ee137905f94f1a1) (#1224) and [`bc1be3d42`](https://github.com/locationtech/jts/commit/bc1be3d420005de2e402bf2e1daa6f8989557626) (#1226).

## Problem
Several call sites computed a direction vector `(dx, dy)` via subtraction before determining its `Quadrant` and/or comparing it for equality. The subtraction can introduce roundoff error, so two direction points that are extremely close (but distinct) can end up comparing as equal / same quadrant when they shouldn't — breaking the total order that `HalfEdge.CompareAngularDirection` (and `IComparable`) are supposed to provide.

## Fix
NTS's `Quadrant` struct already has the endpoint-based constructor `Quadrant(Coordinate p0, Coordinate p1)` (JTS added the equivalent overload in an earlier PR), so this is a call-site-only port:

- `HalfEdge.CompareAngularDirection`: compare `DirectionPt` for equality directly (`Equals2D`) instead of comparing derived `(dx, dy)`, and derive quadrants from `(Orig, DirectionPt)` endpoints instead of `(dx, dy)`.
- `PolygonNodeTopology`: removed the private dx/dy-based `Quadrant(...)` helper, call sites now use `new Quadrant(origin, p)` directly.
- `GeometriesGraph.EdgeEnd.Init`: compute `_quadrant` from `(p0, p1)` *before* computing `_dx`/`_dy`.
- `Planargraph.DirectedEdge` constructor: compute `_quadrant` from `(p0, p1)` before computing `dx`/`dy`.

`GeometriesGraph.DirectedEdge` inherits from `EdgeEnd` and needed no separate change.

## Tests
Ported JTS's regression test as `TestCompareRobust_JTS1224` in `EdgeGraphTest.cs` — adds three near-collinear edges (`(0, 0.5)`, `(0, 0.49999999999999994)`, `(0, 1)`) from a common origin and asserts the two closest direction points don't compare equal and the comparison is antisymmetric.

Verified this test **fails with the old dx/dy-based algorithm** (`eUpper.CompareTo(eLower)` returns `0` instead of a non-zero value) and passes with the fix.

Full `EdgeGraph`/`Planargraph`/`PolygonNodeTopology`/`Overlay`/`RelateNG`/`Noding` test suites: 521 passed, 12 pre-existing skips, 0 failures.

---
**AI disclosure:** This fix was ported and this PR prepared with AI assistance (GitHub Copilot CLI), based on a direct diff against JTS commits a3927e635 and bc1be3d42. Please review before merging.

### DIFF
--- a/src/NetTopologySuite/Algorithm/PolygonNodeTopology.cs
+++ b/src/NetTopologySuite/Algorithm/PolygonNodeTopology.cs
@@ -133,8 +133,8 @@ namespace NetTopologySuite.Algorithm
         /// <returns><c>true</c> if vector P has angle greater than Q</returns>
         private static bool IsAngleGreater(Coordinate origin, Coordinate p, Coordinate q)
         {
-            var quadrantP = Quadrant(origin, p);
-            var quadrantQ = Quadrant(origin, q);
+            var quadrantP = new Quadrant(origin, p);
+            var quadrantQ = new Quadrant(origin, q);
 
             /*
              * If the vectors are in different quadrants, 
@@ -163,8 +163,8 @@ namespace NetTopologySuite.Algorithm
         /// </returns>
         public static int CompareAngle(Coordinate origin, Coordinate p, Coordinate q)
         {
-            var quadrantP = Quadrant(origin, p);
-            var quadrantQ = Quadrant(origin, q);
+            var quadrantP = new Quadrant(origin, p);
+            var quadrantQ = new Quadrant(origin, q);
 
             /*
              * If the vectors are in different quadrants, 
@@ -183,14 +183,6 @@ namespace NetTopologySuite.Algorithm
                 case OrientationIndex.Clockwise: return -1;
                 default: return 0;
             }
-        }
-
-
-        private static Quadrant Quadrant(Coordinate origin, Coordinate p)
-        {
-            double dx = p.X - origin.X;
-            double dy = p.Y - origin.Y;
-            return new Quadrant(dx, dy);
         }
 
     }
--- a/src/NetTopologySuite/EdgeGraph/HalfEdge.cs
+++ b/src/NetTopologySuite/EdgeGraph/HalfEdge.cs
@@ -403,17 +403,13 @@ namespace NetTopologySuite.EdgeGraph
         /// </remarks>
         public int CompareAngularDirection(HalfEdge e)
         {
-            double dx = DirectionX;
-            double dy = DirectionY;
-            double dx2 = e.DirectionX;
-            double dy2 = e.DirectionY;
-
+            // assert: orig() == e.orig()
             // same vector
-            if (dx == dx2 && dy == dy2)
+            if (DirectionPt.Equals2D(e.DirectionPt))
                 return 0;
 
-            var quadrant = new Quadrant(dx, dy);
-            var quadrant2 = new Quadrant(dx2, dy2);
+            var quadrant = new Quadrant(Orig, DirectionPt);
+            var quadrant2 = new Quadrant(e.Orig, e.DirectionPt);
 
             /*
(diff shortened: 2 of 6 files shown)


bjornharrtell:
Same CI issue as PR #870 (`CoverageGapFinderTest.TestThreePolygonGap` failing due to rotation-fragile ring comparison in the test, exposed by this fix's legitimate change to edge ordering). Applied the same fix. CI is now green.

src/NetTopologySuite/Operation/OverlayNG/OverlayEdge.cs
```diff
@@ -102,7 +102,8 @@ public Coordinate[] CoordinatesOriented
         /// <param name="coords">The coordinate list to add to</param>
         public void AddCoordinates(CoordinateList coords)
         {
-            bool isFirstEdge = coords.Count > 0;
+            //-- fixes https://github.com/locationtech/jts/pull/1187 (JTS commit 52c5d9882)
```
AsgerPetersen, line 105: This seems unrelated. Is this PR based on branch from #870 ?

AsgerPetersen (changes requested):
JTS's doc-comment additions to Quadrant.java (noting quadrant(Coordinate,Coordinate) is more robust than quadrant(double,double)) weren't ported. It would be a nice small addition since it documents exactly the rationale this PR relies on

src/NetTopologySuite/Operation/OverlayNG/OverlayEdge.cs
bjornharrtell: Yeah, mistake due to merged here and there. Fixed.

bjornharrtell (commented):
Ported JTS's doc-comment cross-references onto Quadrant(double,double) and Quadrant(Coordinate,Coordinate) noting the latter is more robust. Also rebased onto current develop and dropped the now-redundant duplicate commits (OverlayEdge.cs / CoverageGapFinderTest.cs fixes, both already merged via #870) so the diff is scoped to just the Quadrant/CompareAngularDirection change.

AsgerPetersen (approved):
Thank you. Looks solid now

PR #872, opened 2026-08-28
Add Equals/GetHashCode overrides to value classes for consistency (JTS #1184)

### LINKED ISSUE #866: Add Equals/GetHashCode consistent with CompareTo for LinearLocation, EdgeIntersection, NodeSection, OrientedCoordinateArray (JTS #1184, #1186)

JTS added `equals`/`hashCode` overrides consistent with `compareTo` for several value-semantics `Comparable` classes — [locationtech/jts changelog, 1.21 Bug Fixes](https://github.com/locationtech/jts/blob/master/doc/JTS_Version_History.md) (#1184), plus `DD.hashCode` (#1186, `equals`/`compareTo` side already tracked as N/A under #820 since NTS `DD` is a struct).

Confirmed in current NTS `develop` — none of these override `Equals`/`GetHashCode` (only `IComparable`/`IComparable<T>` is implemented):

- `LinearLocation` — `src/NetTopologySuite/LinearReferencing/LinearLocation.cs`
- `EdgeIntersection` — `src/NetTopologySuite/GeometriesGraph/EdgeIntersection.cs`
- `NodeSection` — `src/NetTopologySuite/Operation/RelateNG/NodeSection.cs`
- `OrientedCoordinateArray` — `src/NetTopologySuite/Noding/OrientedCoordinateArray.cs`

Add `Equals(object)`/`Equals(T)`/`GetHashCode()` overrides to each, consistent with the existing `CompareTo` logic, matching the JTS Java `equals`/`hashCode` implementations.

bjornharrtell:
## Summary

`LinearLocation`, `EdgeIntersection`, `OrientedCoordinateArray` and `NodeSection` implement `IComparable` but did not override `Equals`/`GetHashCode`, so equality was reference-based rather than value-based and inconsistent with `CompareTo`. This can cause subtle bugs when instances are used in hash-based collections (`HashSet`/`Dictionary`) or compared directly.

Ported from JTS commit [`d923a011c`](https://github.com/locationtech/jts/commit/d923a011c75c31d83428a7f63446e955788a9ad2) (locationtech/jts#1201).

## Changes

- `LinearLocation.Equals`/`GetHashCode`: based on componentIndex, segmentIndex, segmentFraction.
- `EdgeIntersection.Equals`/`GetHashCode`: based on segmentIndex, distance.
- `OrientedCoordinateArray.Equals`/`GetHashCode`: orientation-independent, consistent with the existing orientation-independent `CompareTo`.
- `NodeSection.Equals`/`GetHashCode`: based on isA, dimension, id, ringId and edge vertices (assumes same node point, consistent with `CompareTo`).

**Note:** JTS's companion fix to `DD.hashCode` (JTS #1186/locationtech/jts#1202) was **not** ported, since .NET's `double.Equals`/`GetHashCode` already treat `-0.0` and `+0.0` consistently (unlike Java's `Double.equals` vs `==`), so NTS's `DD` struct does not have the same inconsistency.

## Tests

Added regression tests ported/adapted from JTS's `LinearLocationTest`, `OrientedCoordinateArrayTest`, `EdgeIntersectionTest` and `NodeSectionTest`. Full test suite passes (one pre-existing, unrelated failure in `CoverageGapFinderTest` also present on `develop`).

Closes #866

---

**AI disclosure:** This change was implemented with GitHub Copilot CLI assistance, based on a direct diff against the referenced JTS commit and manual verification of the affected NTS classes and their `CompareTo` implementations. Please review carefully before merging.

### DIFF
--- a/src/NetTopologySuite/GeometriesGraph/EdgeIntersection.cs
+++ b/src/NetTopologySuite/GeometriesGraph/EdgeIntersection.cs
@@ -112,5 +112,28 @@ namespace NetTopologySuite.GeometriesGraph
             return _coordinate + " seg # = " + _segmentIndex + " dist = " + _distance;
         }
 
+        /// <summary>
+        /// Tests whether this intersection is at the same location
+        /// (segment index and distance) as another.
+        /// This is consistent with <see cref="CompareTo"/>.
+        /// </summary>
+        /// <param name="obj">The object to compare to</param>
+        /// <returns><c>true</c> if this intersection is at the same location as the argument</returns>
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj == null || GetType() != obj.GetType()) return false;
+            return CompareTo(obj) == 0;
+        }
+
+        /// <inheritdoc cref="object.GetHashCode()"/>
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return (_segmentIndex * 397) ^ _distance.GetHashCode();
+            }
+        }
+
     }
 }
--- a/src/NetTopologySuite/LinearReferencing/LinearLocation.cs
+++ b/src/NetTopologySuite/LinearReferencing/LinearLocation.cs
@@ -362,6 +362,32 @@ namespace NetTopologySuite.LinearReferencing
         }
 
         /// <summary>
+        /// Tests whether this location is equal to another,
+        /// i.e. has the same component index, segment index and segment fraction.
+        /// This is consistent with <see cref="CompareTo(LinearLocation)"/>.
+        /// </summary>
+        /// <param name="obj">The object to compare to</param>
+        /// <returns><c>true</c> if the locations are equal</returns>
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(this, obj)) return true;
+            if (!(obj is LinearLocation other) || GetType() != obj.GetType()) return false;
+            return CompareTo(other) == 0;
+        }
+
+        /// <inheritdoc cref="object.GetHashCode()"/>
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                int hashCode = _componentIndex;
+                hashCode = (hashCode * 397) ^ _segmentIndex;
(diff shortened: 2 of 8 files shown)


fallenmi (changes requested):
Disclosure: I used Codex to assist with this review and verified the finding against exact commit `63f271ca3a4ac6bde6c0e80a42ef99fc4bb09d85`.

Please handle `NaN` segment fractions before defining `LinearLocation.Equals` as `CompareTo(other) == 0`. The existing comparison semantics already treat a `NaN` fraction as equal to an ordinary fraction (the pre-existing `TestZeroLengthLineString` exercises this path), so the new equality is neither transitive nor consistent with its hash code.

On exact head, locations at fractions `0`, `NaN`, and `0.5` produce `zero.Equals(nan) == true`, `nan.Equals(half) == true`, but `zero.Equals(half) == false`. The zero and NaN locations also hash to `0` and `2146435072`, respectively, so a `HashSet<LinearLocation>` containing zero reports `Contains(nan) == false` even though `Equals` says they are equal, and then stores both entries. Exact base does not introduce this value equality, while the current merge has the identical head tree.

Please define explicit NaN semantics that preserve the equality and hash contracts, and add a regression covering this existing NaN location path. The submitted focused tests pass 17/17 and the full suite passes 2,579 with 25 skips, but neither covers this case.

AsgerPetersen (changes requested):
I agree with @fallenmi. CompareTo treats a NaN fraction/distance as equal to any other value (IEEE-754 comparisons with NaN are always false, so it falls through to return 0). That makes the new Equals/GetHashCode non-transitive/inconsistent for LinearLocation and EdgeIntersection. This looks like an upstream JTS bug too (same fallthrough logic in compareTo), but it needs fixing here regardless since NTS is shipping it as a public contract. I took the liberty to add regression tests to the branch.

bjornharrtell:
Nice catch @fallenmi and @AsgerPetersen I've reported upstream https://github.com/locationtech/jts/issues/1228 and proposed a fix https://github.com/locationtech/jts/pull/1229.

bjornharrtell:
@AsgerPetersen upstream bug is fixed and ported here, tests passing. Will merge.

PR #873, opened 2026-08-28
Improve BufferOp robustness for artifacts and invalid coordinates (JTS #1161, #1165)

### LINKED ISSUE #867: BufferOp: add artifact removal heuristic for single-element inputs and handle all-invalid coordinate lists (JTS #1161, #1165)

Two related `BufferOp` robustness fixes from JTS 1.21 that are not present in NTS `develop` (no matching code found under `Operation/Buffer/`):

- [locationtech/jts#1161](https://github.com/locationtech/jts/pull/1161) — buffer artifact removal heuristic for single-element inputs.
- [locationtech/jts#1165](https://github.com/locationtech/jts/pull/1165) — fix `BufferOp` to handle geometries with all-invalid coordinate lists (currently likely throws or produces a bad result instead of a valid/empty buffer).

Port both fixes from JTS `BufferOp`/related buffer classes, along with their JTS regression tests.

bjornharrtell:
## Summary

Two robustness fixes ported from JTS's buffer operation:

### 1. Artifact removal heuristic for single-element inputs (JTS #1161)

Topology robustness problems or buffer curve generation anomalies can occasionally produce spurious small extra polygons ("artifacts") when buffering a geometry. Since a single-element input geometry buffered with `distance > 0` must always produce a single-element result, `BufferBuilder` now keeps only the largest-area polygon when the result has more than one polygon in this case. This does not apply for `distance <= 0`, since `distance = 0` can legitimately produce a multipolygon due to topology collapse, and `distance < 0` may legitimately erode a polygon into disconnected pieces.

Ported from JTS commit [`727943b68`](https://github.com/locationtech/jts/commit/727943b68cc36a9e395c28bfea9b40b720a4f971) (locationtech/jts#1161).

### 2. Handle all-invalid coordinate lists (JTS #1165)

`BufferCurveSetBuilder` could throw/crash when a line, polygon shell, or polygon hole had all of its coordinates removed as invalid (NaN/Infinite) or repeated during cleaning, since it passed the resulting empty coordinate array on to curve generation. Now these elements are skipped instead.

Ported from JTS commit [`fc9ed7dcb`](https://github.com/locationtech/jts/commit/fc9ed7dcb8921e7262633674e817a4f08a4ace25) (locationtech/jts#1165).

## Tests

Added regression tests ported/adapted from JTS's `BufferTest`:
- Artifact removal for line buffers (round and flat end caps)
- All-invalid-coordinate points, lines, polygon shells, and polygon holes

Verified 4 of the 6 new tests fail without the fix (crash, or wrong polygon/geometry count) and pass with it. The other 2 (`TestInvalidCoordPoint_JTS1165`, `TestInvalidCoordHole_JTS1165`) already passed pre-fix due to pre-existing NTS guards not present in JTS (`AddPoint`'s invalid-coordinate check, and `IsRingFullyEroded` treating a degenerate/empty ring as fully eroded) — kept as regression coverage for the ported behavior. Full `Buffer` test suite (167 tests) passes with no regressions.

Closes #867

---

**AI disclosure:** This change was implemented with GitHub Copilot CLI assistance, based on a direct diff against the referenced JTS commits and manual verification against the corresponding NTS buffer code. Please review carefully before merging.

### DIFF
--- a/src/NetTopologySuite/Operation/Buffer/BufferBuilder.cs
+++ b/src/NetTopologySuite/Operation/Buffer/BufferBuilder.cs
@@ -137,8 +137,43 @@ namespace NetTopologySuite.Operation.Buffer
                 return CreateEmptyResultGeometry();
             }
 
+            /*
+             * Heuristic to remove artifacts caused by topology robustness problems
+             * or buffer curve generation anomalies.
+             * Uses fact that for distance > 0 single-element inputs must create single element buffers.
+             * This does not hold if distance <= 0;
+             * distance = 0 can create multipolygon results due to topology collapse,
+             * and distance < 0 may erode polygons so they are disconnected.
+             */
+            if (distance > 0 && g.NumGeometries == 1 && resultPolyList.Count > 1)
+            {
+                resultPolyList = KeepLargestArea(resultPolyList);
+            }
+
             var resultGeom = _geomFact.BuildGeometry(resultPolyList);
             return resultGeom;
+        }
+
+        private static IList<Geometry> KeepLargestArea(IList<Geometry> polyList)
+        {
+            var largest = FindLargestArea(polyList);
+            return new List<Geometry> { largest };
+        }
+
+        private static Geometry FindLargestArea(IList<Geometry> polyList)
+        {
+            double largestArea = 0;
+            Geometry largest = null;
+            foreach (var poly in polyList)
+            {
+                double polyArea = poly.Area;
+                if (largest == null || polyArea > largestArea)
+                {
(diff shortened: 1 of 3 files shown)


AsgerPetersen (approved):
The PR claims that all new tests fail pre-fix. This is not true since two of the tests do not fail without the fix. This is caused by structural differences between JTS and NTS which existed before the fix. I think the tests should be kept, but maybe the PR description should be corrected.

bjornharrtell:
@AsgerPetersen verified and fixed description, thanks!

PR #880, opened 2026-09-01
Fix dx computation in OffsetPointGenerator.ComputeOffsetPoints and add corresponding unit test

### LINKED ISSUE #879: Bug in OffsetPointGenerator.ComputeOffsetPoints
This seems wrong: 

```csharp
double dx = p1.X - p0.Y;
double dy = p1.Y - p0.Y;
```

https://github.com/NetTopologySuite/NetTopologySuite/blob/94b27fc442a5a7ed65b9110378612804b4750c67/src/NetTopologySuite/Operation/Overlay/Validate/OffsetPointGenerator.cs#L81

AsgerPetersen:
Fix incorrect dx computation in OffsetPointGenerator.ComputeOffsetPoints

dx was computed as p1.X - p0.Y instead of p1.X - p0.X — a porting typo, not present in JTS's computeOffsetPoints, which has always used p1.x - p0.x. This rotated the perpendicular offset direction (and thus the "left"/"right" probe points) for any segment whose start point has X != Y.

Impact

This bug currently has no impact on any real overlay result. OverlayResultValidator (the sole consumer of OffsetPointGenerator) is only ever called from SnapIfNeededOverlayOp.cs, and that call is commented out — identically in upstream JTS's SnapIfNeededOverlayOp.java.

Its only other consumer is the JTS/NTS TestRunner's opt-in -geomop/OverlayValidatedGeometryOperation XML test hook — and even JTS's own TestOverlay.xml, the one file that references it, has that tag commented out. No XML test file in either repo currently activates it.

So the bug never corrupted a production overlay result; it only weakened an already-dormant, optional geometric self-check tool. We're fixing it anyway to keep this port faithful to JTS's algorithm, and to avoid latent risk if the validator is ever wired back in (as it is, unused, in JTS itself) or reused elsewhere.

Changes

Fix the dx computation.
Add OffsetPointGeneratorTest covering a non-axis-aligned segment (p0.X != p0.Y), which fails against the old code and passes with the fix.

Closes #879

### DIFF
--- a/src/NetTopologySuite/Operation/Overlay/Validate/OffsetPointGenerator.cs
+++ b/src/NetTopologySuite/Operation/Overlay/Validate/OffsetPointGenerator.cs
@@ -77,7 +77,7 @@ namespace NetTopologySuite.Operation.Overlay.Validate
         /// <param name="offsetPts"></param>
         private void ComputeOffsetPoints(Coordinate p0, Coordinate p1, double offsetDistance, IList<Coordinate> offsetPts)
         {
-            double dx = p1.X - p0.Y;
+            double dx = p1.X - p0.X;
             double dy = p1.Y - p0.Y;
             double len = Mathematics.MathUtil.Hypot(dx, dy);
             // u is the vector that is the length of the offset, in the direction of the segment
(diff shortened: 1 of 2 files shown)


bjornharrtell:
@AsgerPetersen should this not be fixed in upstream first?

AsgerPetersen:
Ah. This is not present in JTS. It is a porting bug introduced when originally porting this code.

bjornharrtell (approved):

bjornharrtell:
@AsgerPetersen ah I misunderstood issue comment.

PR #885, opened 2026-09-18
Fix RelateNG equals for empty geometries

YahorKliutsuk:
### Prerequisites

- [ +] I have written a descriptive pull-request title
- [ +] I have verified that there are no overlapping [pull-requests](https://github.com/NetTopologySuite/NetTopologySuite/pulls) open
- [ +] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository.

- [ +] I have provided test coverage for my change (where applicable)

### Description

`RelateNG.Relate(a, b, RelatePredicate.EqualsTopologically())` returns false for any of two
empty geometries. Upstream returns true and has done since the empty handling was added
to JTS, which was after this part was ported

Two pieces are missing: `RelateNG.Relate` returns early when both inputs are empty, before
either `Init` overload runs, so the predicate never sees the case; upstream has no such
block. And the equals predicate itself doesn't drop the equal-dimension requirement,
doesn't override `requireInteraction()` to false, and doesn't set true when both envelopes
are null

`RelateNGTest.TestEmptyEquals` pins the current behaviour, with a comment noting that
empties didn't test equal in JTS at the time. Upstream's `testEmptyEmpty` now asserts the
opposite, so I've renamed and updated it to match

`Geometry.EqualsTopologically`  doesn't change: it goes through `GeometryRelate.Legacy` by
default, same as JTS. NUnit suite 2620 passing, Vivid 563 passing, both with the CI filter

I haven't run a JTS build to compare, so please sanity-check the upstream claims

Upstream for reference, pinned to `9c995e5`:
* [`RelatePredicate.equalsTopo()`](https://github.com/locationtech/jts/blob/9c995e5383299b6b1596b287871434bf11eabc9f/modules/core/src/main/java/org/locationtech/jts/operation/relateng/RelatePredicate.java#L483-L500) — the three bits above
* [`RelateNG.relate()`](https://github.com/locationtech/jts/blob/9c995e5383299b6b1596b287871434bf11eabc9f/modules/core/src/main/java/org/locationtech/jts/operation/relateng/RelateNG.java#L228-L240) — no early return between building `geomB` and the two `init` calls
* [`RelateNGTest.testEmptyEmpty()`](https://github.com/locationtech/jts/blob/9c995e5383299b6b1596b287871434bf11eabc9f/modules/core/src/test/java/org/locationtech/jts/operation/relateng/RelateNGTest.java#L630-L644)
* [locationtech/jts@6a9af07](https://github.com/locationtech/jts/commit/6a9af07059671bdc6e62542c9739137ab53fd4d8) — the commit that added `misc/TestRelateEmpty.xml`


test/NetTopologySuite.Tests.NUnit/Operation/RelateNG/RelateNGTest.cs
```diff
@@ -723,7 +723,7 @@ public void TestRepeatedPointAA()
 
         //================  Repeated Points  ==============
```
bjornharrtell:
```suggestion
        //================  EMPTY Points  ==============
```

bjornharrtell (changes requested):
Thanks, good catch! LGTM except one minor comment thing, see suggested change.

bjornharrtell (approved):

bjornharrtell:
@jointvirt I took the liberty of commiting the comment change and will merge.

PR #869, opened 2026-08-28
Fix boundary point classification in RelateNG.ComputeLineEnds (JTS #1175)

bjornharrtell:
Closes #862.

Ports JTS commit [`e8de44d9f`](https://github.com/locationtech/jts/commit/e8de44d9fa36049ec165cc308cabaa66bf72957a) (`RelateNG.java`, JTS locationtech/jts#1175).

## Problem
The exterior-intersection optimization in `RelateNG.ComputeLineEnds` used a single `hasExteriorIntersection` flag to decide when it was safe to skip further known-exterior line components. This conflated interior and boundary exterior intersections, causing boundary points of disjoint line components to be incorrectly skipped in some MultiLineString cases, producing wrong `Relate`/DE-9IM results.

## Fix
Track `hasInteriorExteriorIntersection` and `hasBoundaryExteriorIntersection` separately and only short-circuit once **both** are true. `ComputeLineEnd` now returns the endpoint's `Location` (`Interior`/`Boundary`/`Null`) instead of a plain `bool`, matching the JTS fix's return-type change.

## Tests
Ported JTS's regression test (`testLineDisjointMultiLineWithBoundaryInExterior_JTS1175`) as `TestLineDisjointMultiLineWithBoundaryInExterior_JTS1175` in `RelateNGTest.cs`:
```
A: LINESTRING(10 10,20 20)
B: MULTILINESTRING((0 0,1 0),(1 0,2 0),(-1 0,0 0))
Expected IM: FF1FF0102
```
All `RelateNGTest` tests pass (63/63).

Issue #98, opened 2016-03-22
GeoJson error when writing nested objects


DGuidi:
A test built for #58 - [geojson_should_serialize_nested_objects](https://github.com/NetTopologySuite/NetTopologySuite/blob/master/NetTopologySuite.Samples.Console/Tests/Github/Issue58Fixture.cs#L11) - fails. Maybe a regression introduced with latest commits?

PR #886, opened 2026-09-19
Sync XML test corpus with upstream JTS

YahorKliutsuk:
### Prerequisites

- [+] I have written a descriptive pull-request title
- [ +] I have verified that there are no overlapping [pull-requests](https://github.com/NetTopologySuite/NetTopologySuite/pulls) open
- [+] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository.

- [+] I have provided test coverage for my change (where applicable)

### Description
While comparing our XML corpus against JTS I found a few cases that were added or revised
upstream and never made it over here. Nothing dramatic, but they're cheap to carry across,
for each one,the JTS commit it came from:

* `general/TestRelatePA.xml` - three cases with an empty MultiPoint element in A or B
  (locationtech/jts@39fa4a7)
* `general/TestRelateLL.xml` - the case for locationtech/jts#1175; #869 fixed that scenario
  here, so this just pins it (locationtech/jts@532b2ce)
* `misc/TestRelateGC.xml` -two cases that upstream extended to also check the B/A direction
  (locationtech/jts@c3d56e6)
* `failure/TestOverlayNGFailure.xml` - the known OverlayNG failures from locationtech/jts#1000.
  Marked `FailureCase` like its neighbours, so CI keeps skipping it (locationtech/jts@7ff3367)

Since the `general/` fixtures don't report per case, I ran the changed files case by case
locally: RelatePA 121/121, RelateLL 46/46. RelateGC is a bit special - the runner goes
through Legacy relate, which rejects GeometryCollections, so that file wasn't runnable in
CI before either. Evaluated directly with `RelateNG.Relate`, all 300 entries pass.
Vivid suite with the CI filter: 580 tests, 0 failures

Three things I left out on purpose. The new ring cases in `TestBuffer.xml` and two of them
fail here because the matching buffer fix hasn't been ported yet.
I’ll send this separately along with the fix (it’ll be ready soon; I already have a draft of the code. I'll update the description when I make the second pull request). `TestBufferExternal2.xml` #98 — it differs from JTS, but that's
our own adjustment from 7dce5f21, not something upstream changed. And
`misc/TestRelateEmpty.xml`, which needs the runner to evaluate relate through `RelateNG`
first

UPD: 
The buffer fix for the TestBuffer.xml cases mentioned above is in #887.


bjornharrtell (approved):
LGTM, thanks @YahorKliutsuk

PR #887, opened 2026-09-19
Fix buffer of CCW closed lines and rings eroding to empty

YahorKliutsuk:
### Prerequisites

- [+] I have written a descriptive pull-request title
- [+] I have verified that there are no overlapping [pull-requests](https://github.com/NetTopologySuite/NetTopologySuite/pulls) open
- [+] I have verified that I am following the existing coding patterns and practice as demonstrated in the repository.

- [+] I have provided test coverage for my change (where applicable)

### Description
Ran into this while syncing the XML corpus (#886): buffering a closed `LineString` or
`LinearRing` that runs counter-clockwise comes back as `POLYGON EMPTY` as soon as the
distance is big enough to swallow the hole side

    var ring = new WKTReader().Read("LINEARRING (1 9, 1 1, 9 1, 9 9, 1 9)");
    ring.Buffer(10);   // POLYGON EMPTY - should be a rounded square, area ≈ 696

Flip the ring to clockwise and it's fine. Use distance 1 and it's fine too, which is
probably why nobody noticed

What's going on is a double flip. `AddRingBothSides` looks at the ring orientation and
picks a side for each offset curve. It then calls `AddRingSide`, which was written for
polygon rings and does its own orientation check — and flips the side again if the ring
is CCW. So for a CCW ring both curves land on the inside, and once the hole side erodes
there's nothing left

JTS sorted this out in locationtech/jts@6bdb50a (prompted by libgeos/geos#1236) by
separating the two jobs: `AddPolygonRingSide` keeps the CW/CCW handling for polygon rings,
`AddLinearRingSides` (the old `AddRingBothSides`) picks sides on its own and hands them to
a plain `AddRingSide` that doesn't flip anything. This is a straight port of that, plus
its unit test and the three `TestBuffer.xml` ring cases that came with it — the ones I
held back from #886 because two of them failed here

`TestRingCCW` fails on `develop` with `POLYGON EMPTY` and passes with the change. Both
suites with the CI filter: NUnit 2621 / 0 failures, Vivid 563 / 0 failures. `TestBuffer.xml`
case by case: 41/41


bjornharrtell (approved):
LGTM, thanks @YahorKliutsuk

